PR #10, opened 2015-08-18
fix == bashism

vapier:
The test tool only has = for comparison, not ==.  The latter is accepted
by bash and other shells, but is not in POSIX.


pcmoore:
Thanks for the fix, it looks reasonable to me.  Can I add your sign-off to the patch?

vapier:
sorry about that ... should be fixed now

pcmoore:
Thanks.  Fixed in commit 9b3d08d90c40e4fbc4882e0be9585b7673d12a5e

PR #14, opened 2015-08-27
man: seccomp_init: improve kill/trap documentation

vapier:
The current description is fairly light as to the runtime behavior when
the filters fail.  Expand more to make it clear what happens between the
two modes.

Signed-off-by: Mike Frysinger <vapier@gentoo.org>


pcmoore:
Merged in 28e50c6b84aed18cbb886b26879c47f1b2e3ea24, but I had to convert the 'BRm' macro to 'BR' so it would display without error on my system.

Also, in the future, please submit patches via the mailing list.

PR #23, opened 2016-01-01
system.h: Remove conflicting kernel header include

somasis:
This fixes building on musl libc, since musl does not include kernel
headers. I've tested this as working on both glibc and musl.

Signed-off-by: Kylie McClain <somasis@exherbo.org>

### DIFF
--- a/src/system.h
+++ b/src/system.h
@@ -23,7 +23,6 @@
 #define _SYSTEM_H
 
 #include <linux/filter.h>
-#include <linux/prctl.h>
 
 #include "configure.h"
 


anarchpenguin:
This isn't a proper fix either as now you are gonna have used without defines on a glibc system.

somasis:
libseccomp compiled and passed all its tests on a system using glibc-2.22 with linux-headers-4.3, so I'm not entirely sure why it wouldn't be a proper fix.

anarchpenguin:
warning: implicit declaration of function 'prctl' [-Wimplicit-function-declaration]

somasis:
I don't get a single warning compiling with gcc 5.3.0 (on the same glibc system).

anarchpenguin:
/var/tmp/tmpfs/portage/sys-libs/libseccomp-2.2.3/work/libseccomp-2.2.3/src/system.c: In function 'sys_filter_load':
/var/tmp/tmpfs/portage/sys-libs/libseccomp-2.2.3/work/libseccomp-2.2.3/src/system.c:74:8: warning: implicit declaration of function 'prctl' [-Wimplicit-function-declaration]
   rc = prctl(PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0);

That is gcc-5.3.0 glibc-2.22 linux-headers-4.3

anarchpenguin:
You most likely are not passing -Wimplicit-function-declaration hense your failure to see them.

somasis:
[Not a single warning with -Wimplict-function-declaration.](https://gist.github.com/somasis/4aa38eaed4644a755ca4)

anarchpenguin:
ignore my last, I had a bad merge on my tree still sorry for the noise.

anarchpenguin:
somasis please use your original patch it will work as well and be just fine. Less ifdef's make life easier

somasis:
Alright.

anarchpenguin:
Thank you, and once again my apology for the noise. I should have pulled a clean tree and tested it earlier.

somasis:
s'all good

bkuhls:
Positively tested the patch for buildroot: http://patchwork.ozlabs.org/patch/576066/

pcmoore:
Thanks for your help with resolving the problem but I would rather resolve this by using sys/prctl.h instead of removing the include entirely, see faf6919f36eae556f1d4e40b8847070221f553c6 for my proposed fix.

PR #38, opened 2016-05-29
tests: sort out simulation blacklist/whitelist

### LINKED ISSUE #35: BUG: misleading name of test 18-sim-basic_whitelist
Hi,

I was reading [tests/18-sim-basic_whitelist.c](../blob/master/tests/18-sim-basic_whitelist.c).

If I understand it correctly, it does the following:

- Disallow some `read`, `write`, `close`, and `rt_sigreturn` syscalls (only if they act on stdin, stdout, stderr).
- Allow everything else (in particular, reading/writing to any other file descriptor is allowed)

This is not whitelisting, this is blacklisting.

Should the file be renamed? Should all `KILL`s and `ACCEPT`s be swapped to achieve whitelisting?

It would be nice to have a true whitelisting example, since this is the [strongly recommended](http://man7.org/linux/man-pages/man2/seccomp.2.html) use of seccomp.

lucab:
This commit sorts out a mismatch between simulation tests,
properly laying down a blacklist and a whitelist simulation.
Fixes #35

Signed-off-by: Luca Bruno <lucab@debian.org>

### DIFF
--- a/tests/.gitignore
+++ b/tests/.gitignore
@@ -38,3 +38,4 @@ util.pyc
 31-basic-version_check
 32-live-tsync_allow
 33-sim-socket_syscalls_be
+34-sim-basic_blacklist
--- a/tests/18-sim-basic_whitelist.tests
+++ b/tests/18-sim-basic_whitelist.tests
@@ -8,18 +8,18 @@
 test type: bpf-sim
 
 # Testname		Arch	Syscall		Arg0		Arg1		Arg2	Arg3	Arg4	Arg5	Result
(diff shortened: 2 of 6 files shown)


pcmoore:
Thanks for the PR!  If it isn't too much trouble could you please add test 34 to tests/.gitignore?

lucab:
Ah, I didn't spot the nested .gitignore before. Amended.

pcmoore:
Merged in 5e0a33f8f5c086204451041c43010e4ab51b5c6e, thanks!

PR #39, opened 2016-06-10
man: clarify SCMP_FLTATR_CTL_NNP semantic

lucab:
Clarify that the zero value results in a no-op on libseccomp
side, and applications will need to have to have proper caps
or set NO_NEW_PRIVS by themself.

Signed-off-by: Luca Bruno <lucab@debian.org>


lucab:
I stepped into this while exploring a more complex systemd scenario, and I was not sure about proper usage. I digged into source to be sure about the semantic, and I think re-wording the manpage in this way makes it more explicit.

/cc @alban

pcmoore:
Merged in 937e774d8ccfd063a601f49ef46769de3926b62d, thanks!

Issue #40, opened 2016-06-10
strange behaviour on v4.4 kernels with libseccomp v2.3.1


xnox:
Hello,

I'm in a i686 chroot, and using scmp_sys_resolver I see multiple regression with syscall look-ups. The tests I run is to check that number->name->number lookups are roundtrip safe, and they appear to not be as per below:
```
# ./scmp_sys_resolver 373
shutdown
# ./scmp_sys_resolver shutdown
-113
# ./scmp_sys_resolver '\-113'
-1
# ./scmp_sys_resolver -t shutdown
102
# ./scmp_sys_resolver 102
socketcall
```

My expectation was for the output to just be:
```
# ./scmp_sys_resolver 373
shutdown
# ./scmp_sys_resolver shutdown
373
```

I see this regression for other syscalls too, on i686 and on s390x architectures.
amd64 and ppc64el pass correctly.

s390x failures:
```
FAIL: 357 (recvmmsg) != recvmmsg (-119)
FAIL: 358 (sendmmsg) != sendmmsg (-120)
FAIL: 359 (socket) != socket (-101)
FAIL: 360 (socketpair) != socketpair (-108)
FAIL: 361 (bind) != bind (-102)
FAIL: 362 (connect) != connect (-103)
FAIL: 363 (listen) != listen (-104)
FAIL: 364 (accept4) != accept4 (-118)
FAIL: 365 (getsockopt) != getsockopt (-115)
FAIL: 366 (setsockopt) != setsockopt (-114)
FAIL: 367 (getsockname) != getsockname (-106)
FAIL: 368 (getpeername) != getpeername (-107)
FAIL: 369 (sendto) != sendto (-111)
FAIL: 370 (sendmsg) != sendmsg (-116)
FAIL: 371 (recvfrom) != recvfrom (-112)
FAIL: 372 (recvmsg) != recvmsg (-117)
FAIL: 373 (shutdown) != shutdown (-113)
```

i686 failures:
```
FAIL: 337 (recvmmsg) != recvmmsg (-119)
FAIL: 345 (sendmmsg) != sendmmsg (-120)
FAIL: 359 (socket) != socket (-101)
FAIL: 360 (socketpair) != socketpair (-108)
FAIL: 361 (bind) != bind (-102)
FAIL: 362 (connect) != connect (-103)
FAIL: 363 (listen) != listen (-104)
FAIL: 364 (accept4) != accept4 (-118)
[...]

PR #41, opened 2016-06-10
Drop table overrides in x86_syscall_resolve_name() to pseudo syscall …

xnox:
…numbers

A bunch of syscalls got converted from pseudo to real syscall
numbers. And the table now corectly uses pseudo syscall numbers where
needed (e.g. accept, send) thus there is no need to repeat these
overrides in the x86_syscall_resolve_name() which has now got out of
sync with the table.

Fixes https://github.com/seccomp/libseccomp/issues/40

Signed-off-by: Dimitri John Ledkov <xnox@ubuntu.com>

### DIFF
--- a/src/arch-x86-syscalls.c
+++ b/src/arch-x86-syscalls.c
@@ -468,49 +468,6 @@ int x86_syscall_resolve_name(const char *name)
 	unsigned int iter;
 	const struct arch_syscall_def *table = x86_syscall_table;
 
-	/* XXX - plenty of room for future improvement here */
-
-	if (strcmp(name, "accept") == 0)
-		return __PNR_accept;
-	if (strcmp(name, "accept4") == 0)
-		return __PNR_accept4;
-	else if (strcmp(name, "bind") == 0)
-		return __PNR_bind;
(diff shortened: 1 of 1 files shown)


xnox:
After this, my tests as pointed out in #40 pass, however, the upstream test-suite now fails with:

Test 30-sim-socket_syscalls%%012-00001 result:   FAILURE bpf_sim resulted in ALLOW

xnox:
Horum.

```
30-sim-socket_syscalls  +x86    accept4         0               1               2       N       N       N       KILL
```
Not sure why it should be a KILL and/or how&why my code change results in this test suite failure, when e.g. "accept" ALLOW/KILL tests remain operating correctly.

xnox:
Actually, i"m not sure if this makes sense anymore.

accept4 is 364 on x86 from now on. Thus these two tests are identical, yet expect different results
```
30-sim-socket_syscalls  +x86    364             0               1               2       N       N       N       ALLOW
30-sim-socket_syscalls  +x86    accept4         0               1               2       N       N       N       KILL
```
which is not possible....

xnox:
actually this is intentional.

Issue #42, opened 2016-07-08
Q: clarify the seccomp_syscall_resolve_name_rewrite() behavior


lucab:
I was reading the doc and the source in parallel, and I'm still not sure about the behavior and recommended usage of `seccomp_syscall_resolve_name_rewrite()`.

According to the manpage:

```
[seccomp_syscall_resolve_name*()] resolve the commonly used syscall
name to the syscall number used by the kernel and the rest of the libseccomp API,
with seccomp_syscall_resolve_name_rewrite() rewriting the syscall number for
architectures that modify the syscall.
```

Some doubts I have:
- How does re-writing interact with pseudo-syscalls? Perl bindings are [a bit more explicit](http://search.cpan.org/~mgv/Linux-Seccomp-0.001/lib/Linux/Seccomp.pm) here, but I'm not sure if correct and cannot reproduce that behavior on x86_64. I'll be happy to get some hints/pointer/examples here.
- Which name resolving function should be recommended and used by consumers? "_rewrite" seems to be more precise in some cases (which?), but is the non-rewrite behavior a concern people should be aware? Does it have security implications?
- The default/simpler `seccomp_syscall_resolve_name` has a non-rewrite behavior. Is that on purpose and what are the implications? Would it make sense to switch it to the rewrite behavior (and if not, why)?
- When resolving with a non-rewrite behavior, will other libseccomp functions perform the rewriting internally or not?

Sorry for the long list of questions, but I didn't get a clear answer for this when navigating through the code. I think I got the general idea but I fear that some details in it may be wrong, so I'd prefer if they could be clarified here from a knowledgeable source. I'll be happy to expand the docs (if needed) with those inputs.

pcmoore:
Let me try to answer all your questions, point by point.  Let me know if anything is still confusing at the end.  In the future you might try directing questions like these to the mailing list:

* https://groups.google.com/forum/#!forum/libseccomp

> How does re-writing interact with pseudo-syscalls? Perl bindings are a bit more explicit here, but I'm not sure if correct and cannot reproduce that behavior on x86_64. I'll be happy to get some hints/pointer/examples here.

Good question.  Basically rewriting only happens when the syscall name in question resolves to a pseudo syscall, e.g. the ipc() syscalls on 32-bit x86.  Keep in mind that syscall rewriting is relatively rare, with only 32-bit x86 and s390/s390x providing syscall rewriting routines.

[...]

PR #43, opened 2016-07-11
RFE: clarify syscall number rewriting

lucab:
In case of multiplexed syscalls, syscall name resolver and rule builder
both offer additional functions to ignore or perform syscall number
rewriting.
This commit introduces additional explicit details to the corresponding
manpages.

Signed-off-by: Luca Bruno lucab@debian.org
Closes: #42


pcmoore:
Merged via a66f000de92c7ced17ae6615837dabafd803ae42.

PR #52, opened 2016-11-03
RFE: removed unnecessary call in `seccomp_add_arch`

guoger:
Some logic in the implementation of `seccomp_add_arch` can never
be reached and `arch_def_lookup` was called redundantly.

Signed-off-by: Jiannan Guo <guojiannan1101@gmail.com>


pcmoore:
This looks good - thanks!

pcmoore:
Merged via 75da81e9ae7d5995f0a44673d4a737dafc05c11a.

PR #63, opened 2017-01-02
RFE: utility for resolving architectures

tyhicks:
This pull request adds an (uninstalled) utility for resolving architecture tokens into names suitable for reading.

### DIFF
--- a/tools/Makefile.am
+++ b/tools/Makefile.am
@@ -31,12 +31,13 @@ noinst_PROGRAMS = \
 
 EXTRA_DIST = check-syntax scmp_app_inspector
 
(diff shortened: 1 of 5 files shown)


pcmoore:
Without looking at the code too closely, one thing jumps out at me: the libseccomp arch token is not the same as the "arch" field reported by the kernel's audit records, see x86_64/x32 as an example.

pcmoore:
Okay, I'm going to close this out without merging because of the reasons listed above.  If you really want to translate the architecture token in the audit messages, I might suggest using the ausearch tool with the '-i' parameter.

PR #64, opened 2017-01-02
RFE: add support for SECCOMP_RET_AUDIT

tyhicks:
This pull request adds support for SECCOMP_RET_AUDIT which is a newly proposed seccomp return action that audits the syscall before allowing it to happen. It is similar to SECCOMP_RET_ALLOW in that the syscall is allowed but differs in that the syscall will be audited.

### DIFF
--- a/include/seccomp.h.in
+++ b/include/seccomp.h.in
@@ -256,6 +256,10 @@ struct scmp_arg_cmp {
  */
 #define SCMP_ACT_TRACE(x)	(0x7ff00000U | ((x) & 0x0000ffffU))
 /**
+ * Allow the syscall to be executed after an audit message is emitted
+ */
+#define SCMP_ACT_AUDIT		0x7ffe0000U
+/**
  * Allow the syscall to be executed
  */
 #define SCMP_ACT_ALLOW		0x7fff0000U
--- a/src/db.c
+++ b/src/db.c
@@ -652,6 +652,8 @@ int db_action_valid(uint32_t action)
 		return 0;
 	else if (action == SCMP_ACT_TRACE(action & 0x0000ffff))
 		return 0;
+	else if (action == SCMP_ACT_AUDIT)
+		return 0;
 	else if (action == SCMP_ACT_ALLOW)
 		return 0;
 
--- a/src/gen_pfc.c
+++ b/src/gen_pfc.c
@@ -129,6 +129,9 @@ static void _pfc_action(FILE *fds, uint32_t action)
 	case SCMP_ACT_TRACE(0):
 		fprintf(fds, "action TRACE(%u);\n", (action & 0x0000ffff));
 		break;
+	case SCMP_ACT_AUDIT:
+		fprintf(fds, "action AUDIT;\n");
+		break;
 	case SCMP_ACT_ALLOW:
 		fprintf(fds, "action ALLOW;\n");
 		break;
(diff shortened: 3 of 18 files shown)


pcmoore:
Thanks for the PR.  I'm going to defer looking at this until we see if/when the kernel functionality lands upstream.  However, one thing to think about, and my apologies if you've already taken this into consideration in this PR, how do we detect if SECCOMP_RET_AUDIT is supported at runtime?

tyhicks:
On 01/02/2017 03:59 PM, Paul Moore wrote:
> Thanks for the PR. I'm going to defer looking at this until we see
> if/when the kernel functionality lands upstream. However, one thing to
> think about, and my apologies if you've already taken this into
> consideration in this PR, how do we detect if SECCOMP_RET_AUDIT is
> supported at runtime?

I considered the need for runtime detection of SECCOMP_RET_AUDIT and
decided that it likely wouldn't be a requirement for two reasons:

1) The seccomp kernel code fails close to SECCOMP_RET_KILL when the
return action is unrecognized.

2) Runtime detection wasn't added for the kill, trap, or errno actions
(commit 308ad5f).

Do you consider runtime detection for SECCOMP_RET_AUDIT to be a
requirement? If so, it'll likely drive changes in the kernel patch set.

pcmoore:
You've probably already seen this, but if you haven't, we do have some detection hacks in src/system.c ... look at sys_chk_seccomp_syscall().

tyhicks:
>      1. The seccomp kernel code fails close to SECCOMP_RET_KILL when the
>         return action is unrecognized.
> 
> While this may be the "safe" option, it seems unlikely that application
> developers like converting an AUDIT action into a KILL action. I can
> already see the bug reports and hate mail flowing into my inbox ...

Hehe, good point. :)

> 
>      2. Runtime detection wasn't added for the kill, trap, or errno
>         actions (commit 308ad5f
>         <https://github.com/seccomp/libseccomp/commit/308ad5f1ef3efc0437bd428af4ff116c740b2754>).
> 
> Look again at when that commit landed in libseccomp. Those actions have
> been supported by libseccomp since our first tagged release (v0.1.0) and
> were supported by the current kernel at that time. Since there was no
> libseccomp "legacy" at that point it was reasonable to require certain
> things, now that libseccomp is a bit more mature and is in use by a
> number of projects, we have to be more careful about older kernel versions.

That makes sense.

> 
>     Do you consider runtime detection for SECCOMP_RET_AUDIT to be a
>     requirement? If so, it'll likely drive changes in the kernel patch set.
> 
> We need some way to safely handle SECCOMP_RET_AUDIT on older kernels
> that are not aware of this action; that's the requirement, and runtime
> detection is one way to solve that (we can safely avoid using AUDIT on
> kernels where it is not supported), but there may be others. Don't be
> afraid to get clever if needed, just remember that the code can't be too
> ugly, I've got to maintain this thing for a while ;)

Understood. I'll see what I can come up with.

pcmoore:
@tyhicks what is the status of the kernel patches?  I thought you were planning on submitting a revision taking into account the feedback you received, but I haven't seen anything; is this still on your todo list?

tyhicks:
@pcmoore I'm still working on the second revision. A couple weeks travel and another week of catchup starved those patches for time but look for the patch set to go out later this week or early next.

pcmoore:
@tyhicks Great, no rush, I just wanted to make sure someone was still working on this; I'm happy to hear you are following up with a second revision.

pcmoore:
I think we can close this PR since it isn't relevant anymore, right @tyhicks?

tyhicks:
Yeah, I should just create an all new PR. Please close this one. Thanks!On Aug 17, 2017 12:44 PM, Paul Moore <notifications@github.com> wrote:I think we can close this PR since it isn't relevant anymore, right @tyhicks?

PR #90, opened 2017-08-17
man: Fix SCMP_FLTATR_API_TSKIP typo in seccomp_attr_set man page

tyhicks:
Fix a minor but potentially confusing man page typo.

PR #91, opened 2017-08-24
tests: fix conditional that was skipping all basic python tests

tyhicks:
A conditional added in ec6f45ab was incorrectly comparing the (empty)
stdout of grep -q against 0, which always evaluated to be true and
skipped the basic python tests.

Fix it by using bash's pattern matching.

Signed-off-by: Tyler Hicks <tyhicks@canonical.com>

### DIFF
--- a/tests/regression
+++ b/tests/regression
@@ -608,7 +608,7 @@ function run_test_basic() {
 	local cmd
 
 	# if the test is a script, only run it in native/c mode
-	if [[ $mode != "c" && $(echo "$2" | grep -q '.sh$') -eq 0 ]]; then
+	if [[ $mode != "c" && "$2" == *.sh ]]; then
 		print_result "$1" "SKIPPED" "(only valid in native/c mode)"
 		stats_skipped=$(($stats_skipped+1))
 		return


tyhicks:
As mentioned in #90, the CI tests fail here due to a timeout:

`The job exceeded the maximum time limit for jobs, and has been terminated.`

When running the tests locally without and then with this change, I see that the four basic python tests are ran and succeed. Those tests are 11-basic-basic_errors.py, 13-basic-attrs.py, 15-basic-resolver.py, and 31-basic-version_check.py.

pcmoore:
This seems to work for everything except for test 38-basic-pfc_coverage, which is still skipped:

    ./regression -b 38-basic-pfc_coverage
    =============== Tue Sep 12 18:37:27 EDT 2017 ===============
    Regression Test Report ("regression -b 38-basic-pfc_coverage")
     batch name: 38-basic-pfc_coverage
     test mode:  c
     test type:  basic
    Test 38-basic-pfc_coverage%%001-00001 result:   SUCCESS
     batch name: 38-basic-pfc_coverage
     test mode:  python
     test type:  basic
    Test 38-basic-pfc_coverage%%001-00001 result:   SKIPPED (only valid in native/c mode)
    Regression Test Summary
     tests run: 1
     tests skipped: 1
     tests passed: 1
     tests failed: 0
     tests errored: 0

I haven't done much debugging on this yet, but wanted to mention it in case you ran across something like this after the PR was posted.

pcmoore:
Ah, wait, nevermind, my mistake ... this looks fine, I was just a bit rusty and looking at this wrong, sorry for the confusion.

pcmoore:
Merged via b50312919d2c5f7fe9ecc526f783e8835aed1fc9

PR #98, opened 2017-10-09
BUG: API level fixups

tyhicks:
Here are some minor fixups that I identified while reviewing the API level commit (e89d18205c7dcd7582f41051cd6389c9b12dfccf).

### DIFF
--- a/tests/39-basic-api_level.c
+++ b/tests/39-basic-api_level.c
@@ -40,5 +40,21 @@ int main(int argc, char *argv[])
 	if (api != 1)
 		return -3;
 
+	rc = seccomp_api_set(2);
+	if (rc != 0)
+		return -4;
+	api = seccomp_api_get();
+	if (api != 2)
+		return -5;
+
+	/* Attempt to set a high, invalid API level */
+	rc = seccomp_api_set(1024);
+	if (rc != -EINVAL)
+		return -6;
+	/* Ensure that the previously set API level didn't change */
+	api = seccomp_api_get();
+	if (api != 2)
+		return -7;
+
 	return 0;
 }
--- a/tests/39-basic-api_level.py
+++ b/tests/39-basic-api_level.py
@@ -4,7 +4,9 @@
 # Seccomp Library test program
 #
 # Copyright (c) 2016 Red Hat <pmoore@redhat.com>
-# Author: Paul Moore <paul@paul-moore.com>
+# Copyright (c) 2017 Canonical Ltd.
+# Authors: Paul Moore <paul@paul-moore.com>
+#          Tyler Hicks <tyhicks@canonical.com>
 #
 
 #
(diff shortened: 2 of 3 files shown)


tyhicks:
FYI, I'm pretty sure that https://github.com/seccomp/libseccomp/pull/98/commits/8f38116ca81c2f3e676301207651bee8867f4e7f is the last commit that I'll push to this branch. I can't spot any other issues.

tests/39-basic-api_level.c
```diff
@@ -28,6 +28,7 @@ int main(int argc, char *argv[])
 {
 	int rc;
 	unsigned int api;
+	const unsigned int max_api = 2;
```
pcmoore: Let's bump the `max_api` value up to something absurdly high, 1024 for example, and only test to make sure it is flagged as invalid.  Otherwise I fear this test is going to be a bit too fragile.

tests/39-basic-api_level.c
```diff
@@ -40,5 +41,20 @@ int main(int argc, char *argv[])
 	if (api != 1)
 		return -3;
 
+	rc = seccomp_api_set(max_api);
```
pcmoore: See the comment above, let's drop this check.

As odd as it may sound, I'm okay with putting in explicit checks for each supported API level as the API level is added to the library.

tyhicks: I should have left a comment in the code with my intentions but I was thinking that this would force anyone adding a new API level to add a set/get test for it.

tests/39-basic-api_level.py
```diff
@@ -28,8 +30,32 @@
...
+    if api != 1:
+        raise RuntimeError("Failed getting API level 1")
+
+    set_api(2)
+    api = get_api()
+    if api != 2:
+        raise RuntimeError("Failed getting API level 2")
+
+    try:
+        set_api(3)
```
pcmoore: Similar to my comments in the C variant of this test, let's bump this up to something really large, e.g. 1024.

pcmoore:
My experiment at marking individual commits as "approved" failed miserably, so here are the patches I merged:

* a1c4094f8 api: Handle all possible return values when checking flag support
* 88f883253 system: Add missing param to sys_chk_seccomp_flag() comment
* 6ee8df168 tests: Add new API level test binary to gitignore

The other two commits look pretty good, I just had a few small comments (made inline, if things got mangled let me know and I'll just repeat the comments here).

pcmoore:
Oh, I forgot to add my thanks for cleaning up my mess! :)

tests/39-basic-api_level.c
pcmoore: Yes, I understand, and generally applaud tests like that, but I'm thinking that any new API level addition is going to require additional tests anyway so it isn't as large a concern.  I guess I could be convinced that this is the Right Thing To Do if you feel strongly about it.

tyhicks: I don't feel strongly about it. I just wanted to make sure you knew what I was thinking when I wrote the tests since I did a poor job of recording what I was thinking. I'll get this PR updated hopefully by EOD. Thanks!

tyhicks:
I've updated the C and Python test as you requested. I did away with the "max_api" variable in the C test for two reasons:

1. It didn't exist in the Python test and that bothered me a bit.
1. It is no longer the "max api" in the sense that it is the highest valid API level so the name no longer made sense.

> My experiment at marking individual commits as "approved" failed miserably, ...

As a fellow patch-and-mailing-list dinosaur, I feel your pain. :)

(though, I'm warming up to this new world)

> Oh, I forgot to add my thanks for cleaning up my mess! :)

Pfft. Thanks for helping to clean up all of my messes spread across both PRs. :)

pcmoore:
Thanks for the quick turnaround on the updates, everything looks good and should now be in the master branch:

* 4f16fe208 python: Expose API level functionality
* f9d757de2 tests: Improve seccomp_api_set() test coverage

PR #99, opened 2017-10-12
BUG: minor python pydoc fixes

skrap:


skrap:
(First time contributing to the project — hopefully I'm doing this properly!)

src/python/seccomp.pyx
```diff
@@ -56,13 +56,17 @@ Example:
...
+    f.add_rule(ALLOW, "exit_group")
+    f.add_rule(ALLOW, "brk")
+
     # add syscall filter rules to allow certain syscalls
     f.add_rule(ALLOW, "open")
     f.add_rule(ALLOW, "close")
-    f.add_rule(ALLOW, "read", Arg(0, EQ, sys.stdin))
-    f.add_rule(ALLOW, "write", Arg(0, EQ, sys.stdout))
-    f.add_rule(ALLOW, "write", Arg(0, EQ, sys.stderr))
-    f.add_rule(ALLOW, "rt_sigreturn")
```
tyhicks: Why is `rt_sigreturn` removed here?

pcmoore:
Welcome to libseccomp @skrap and thanks for the patches :)  I have the same question as @tyhicks regarding `rt_sigreturn`, could you please explain that change?  Also as Tyler pointed out, please take a closer look at CONTRIBUTING.md, thanks!

skrap:
You know, I took it out because it wasn't needed on my platform to complete a trivial python program, whereas the others were.  I suppose it could be glibc differences, or python differences... or maybe rt_sigreturn should be added back?  What do you think?

It's actually pretty hard to find a minimal yet useful set of seccomp filters!

tyhicks:
I think that if `rt_sigaction` is required, `rt_sigreturn` should also be included to return from the signal handler that was configured by `rt_sigaction`. Your small test program is likely getting "lucky" in that a signal is not being delivered during the program's short lifetime.

I think there's two things left for you to do to get these patches merged:
1. Update the second patch to leave `rt_sigreturn` in the example
 - You probably should move that add_rule() up to the block of minimum required system calls that you added
1. Update both the first and second patch to include your signed-off-by, as documented in CONTRIBUTING.md

pcmoore:
I agree with @tyhicks, it's best to leave `rt_sigaction` in the syscall list.  I would simply add the `exit_group`, `brk`, and `rt_sigaction` to the existing block of syscalls; there is no need to put them in a separate block.  If I wanted to be picky, I would suggest the following order:

* open
* close
* read
* write
* brk
* rt_sigreturn
* rt_sigaction
* exit_group

Beyond that all you need to do is follow the guidance in CONTRIBUTING.md as previously discussed.

drakenclimber:
I looked through the changes (and ran them, too.)  I still think @tyhicks comment about re-adding `rt_sigreturn` is correct.  What makes you think it isn't needed?  I wasn't able to find a definitive comment why it can be safely removed.

Other than that, I think it looks good.

skrap:
> I still think @tyhicks comment about re-adding `rt_sigreturn` is correct.

@drakenclimber: Fixed, thanks for the feedback!

drakenclimber:
Thanks!  Looks good to me.

pcmoore:
Merged via 853a24c8720d64ef1dfe0f69202f19bf916ddccb - thanks everyone!

PR #108, opened 2018-02-20
RFE: add RISC-V support

Icenowy:


src/arch-riscv64-syscalls.c
```diff
@@ -0,0 +1,528 @@
+/**
+ * Enhanced Seccomp AArch64 Syscall Table
```
shlevy: Copy-paste error?

pcmoore:
Thanks @Icenowy, do you know the current status of the kernel support?

For reference, I just created issue #110 to track the different PRs.

src/arch-riscv64.h
```diff
@@ -0,0 +1,36 @@
+/**
+ * Enhanced Seccomp AArch64 Syscall Table
```
shlevy: Ditto

src/arch-syscall-validate
```diff
@@ -376,6 +376,47 @@ function dump_lib_ppc64() {
 	dump_lib_arch ppc64
 }
 
+#
+# Dump the aarch64 system syscall table
```
shlevy: Ditto

src/arch.c
```diff
@@ -152,6 +155,8 @@ const struct arch_def *arch_def_lookup(uint32_t token)
 		return &arch_def_ppc64;
 	case SCMP_ARCH_PPC64LE:
 		return &arch_def_ppc64le;
+	case SCMP_ARCH_RISCV64:
+		return &arch_def_ppc64;
```
shlevy: Ditto

shlevy:
Awesome! I have a use case that needs to block setuid binary creation, I'll hopefully test that in qemu-system and qemu-user this week (hardware is coming in a few months :smile: )

sorear:
@pcmoore Kernel support is upstream in 4.15.0.  (More drivers are needed to boot, they are expected in 4.17, but 4.15 has the complete uapi)

pcmoore:
> Kernel support is upstream in 4.15.0 ...

@sorear I'm looking at Linus' tree right now (v4.16-rc2+) and I don't see `HAVE_ARCH_SECCOMP_FILTER` in the RISC-V Kconfigs, what am I missing?

pcmoore:
@Icenowy I haven't looked closely, but it looks like you are missing a lot of the test bits needed to verify the RISC-V support.

sorear:
> I'm looking at Linus' tree right now (v4.16-rc2+) and I don't see HAVE_ARCH_SECCOMP_FILTER in the RISC-V Kconfigs, what am I missing?

Ah, sorry, I thought you meant the port in general.  I don't know about seccomp status.

pcmoore:
@sorear while the general Linux RISC-V support is important, for the purposes of libseccomp, if the `HAVE_ARCH_SECCOMP_FILTER` support is not there for RISC-V then there is nothing we can do.

If you, or someone else, is planning on adding the Linux Kernel support soon we can consider reviewing this PR and possibly merging it into a development branch if it would help you, but if the kernel support is not in progress I'm going to recommend we close this PR.

*Please note the closing this PR doesn't mean I'm rejecting the idea of libseccomp on RISC-V, it's just the opposite, I would be very happy to add RISC-V support!  See #110 for the tracking issue.*

pcmoore:
Looking at Linus' tree today, it appears that kernel support is still missing, I'm going to close this PR, please reopen or resubmit when the kernel has the necessary seccomp-bpf support (see above).

PR #109, opened 2018-02-20
api: Add support for SCMP_ACT_KILL_PROCESS

drakenclimber:
This patch addresses issue 96 - "RFE: add support for SECCOMP_RET_KILL_PROCESS" and issue 93 - "BUG: compatibility with openmp".

Test results w/ code coverage:

```./configure --enable-code-coverage
make check-code-coverage

Regression Test Summary
tests run: 7895 
tests skipped: 38
tests passed: 7895 
tests failed: 0
tests errored: 0
PASS: regression

Line Coverage
91.0%
2160 / 2374 
(prior to these changes, code coverage was 2153 / 2367)

Functions
90.3%
140 / 155


Test results w/ python enabled:

./configure --enable-python
make check

Regression Test Summary
 tests run: 15756
 tests skipped: 110
 tests passed: 15756
 tests failed: 0
 tests errored: 0
PASS: regression
```

### DIFF
--- a/src/gen_pfc.c
+++ b/src/gen_pfc.c
@@ -117,8 +117,11 @@ static void _pfc_arg(FILE *fds,
  */
 static void _pfc_action(FILE *fds, uint32_t action)
 {
-	switch (action & 0xffff0000) {
-	case SCMP_ACT_KILL:
+	switch (action & SECCOMP_RET_ACTION_FULL) {
+	case SCMP_ACT_KILL_PROCESS:
+		fprintf(fds, "action KILL-PROCESS;\n");
+		break;
+	case SCMP_ACT_KILL_THREAD:
 		fprintf(fds, "action KILL;\n");
 		break;
 	case SCMP_ACT_TRAP:
--- a/src/gen_pfc.c
+++ b/src/gen_pfc.c
@@ -35,6 +35,7 @@
 #include "db.h"
 #include "gen_pfc.h"
 #include "helper.h"
+#include "system.h"
 
 struct pfc_sys_list {
 	struct db_sys_list *sys;
@@ -110,10 +111,6 @@ static void _pfc_arg(FILE *fds,
 		fprintf(fds, "$a%d", node->arg);
 }
 
-#ifndef SECCOMP_RET_ACTION_FULL
-#define SECCOMP_RET_ACTION_FULL 0xffff0000U
-#endif
-
 /**
  * Display a string representation of the filter action
  * @param fds the file stream to send the output
@@ -123,7 +120,7 @@ static void _pfc_action(FILE *fds, uint32_t action)
 {
 	switch (action & SECCOMP_RET_ACTION_FULL) {
 	case SCMP_ACT_KILL_PROCESS:
-		fprintf(fds, "action KILL-PROCESS;\n");
+		fprintf(fds, "action KILL_PROCESS;\n");
 		break;
 	case SCMP_ACT_KILL_THREAD:
 		fprintf(fds, "action KILL;\n");
--- a/src/gen_pfc.c
+++ b/src/gen_pfc.c
@@ -110,6 +110,10 @@ static void _pfc_arg(FILE *fds,
 		fprintf(fds, "$a%d", node->arg);
 }
 
+#ifndef SECCOMP_RET_ACTION_FULL
+#define SECCOMP_RET_ACTION_FULL 0xffff0000U
+#endif
+
 /**
  * Display a string representation of the filter action
  * @param fds the file stream to send the output
--- a/src/system.c
+++ b/src/system.c
@@ -43,6 +43,7 @@ static int _support_seccomp_syscall = -1;
 static int _support_seccomp_flag_tsync = -1;
 static int _support_seccomp_flag_log = -1;
 static int _support_seccomp_action_log = -1;
+static int _support_seccomp_kill_process = -1;
 
 /**
  * Check to see if the seccomp() syscall is supported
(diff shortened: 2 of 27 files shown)


src/system.c
```diff
@@ -123,7 +123,26 @@ void sys_set_seccomp_syscall(bool enable)
...
-	if (action == SCMP_ACT_KILL) {
+	int rc, nr_seccomp;
+	uint32_t kaction = action & SECCOMP_RET_ACTION_FULL;
+
+
+	nr_seccomp = arch_syscall_resolve_name(arch_def_native, "seccomp");
+	if (nr_seccomp != __NR_SCMP_ERROR) {
+		rc = syscall(nr_seccomp, SECCOMP_GET_ACTION_AVAIL, 0,
+			     &kaction);
+		if (rc == -EPERM)
```
drakenclimber: I'm not fond of this "if", but I couldn't figure out a more graceful solution.  We really should ask the kernel if the requested action is supported here, but in Valgrind's case, the seccomp syscall [isn't supported](http://repo.or.cz/valgrind.git/blob/6ae2edea014669d8082747f0f268e9404e0fd296:/coregrind/m_syswrap/syswrap-amd64-linux.c#l840).

I tried using sys_chk_seccomp_syscall(), but that will fail in a few tests because _support_seccomp_syscall gets set to 1 when the api level is set.  But, _nr_seccomp isn't set using that interface and remains -1.  Test 06-sim-actions.c exhibits this behavior.

drakenclimber: Kees had also mentioned using `/proc/sys/kernel/seccomp/actions_avail` as another avenue to get the available actions.  But that produces a list like the following - `kill_process kill_thread trap errno trace log allow
`, and I wasn't keen on doing string processing.

drakenclimber:
Bah!  I'll look into the CI failure.

src/gen_pfc.c
```diff
@@ -117,8 +117,11 @@ static void _pfc_arg(FILE *fds,
  */
 static void _pfc_action(FILE *fds, uint32_t action)
 {
-	switch (action & 0xffff0000) {
-	case SCMP_ACT_KILL:
+	switch (action & SECCOMP_RET_ACTION_FULL) {
+	case SCMP_ACT_KILL_PROCESS:
+		fprintf(fds, "action KILL-PROCESS;\n");
```
pcmoore: Nitpick: since the #define uses an underscore, let's use an underscore here:

    fprintf(fds, "action KILL_PROCESS;\n");

src/system.c
```diff
@@ -123,7 +123,26 @@ void sys_set_seccomp_syscall(bool enable)
  */
 int sys_chk_seccomp_action(uint32_t action)
 {
-	if (action == SCMP_ACT_KILL) {
+	int rc, nr_seccomp;
+	uint32_t kaction = action & SECCOMP_RET_ACTION_FULL;
```
pcmoore: We need to prepare ourselves for building on system that doesn't define `SECCOMP_RET_ACTION_FULL`; it didn't exist before Kees' patch in August 2017.

src/system.c
```diff
@@ -123,7 +123,26 @@ void sys_set_seccomp_syscall(bool enable)
...
+
```
pcmoore: One line of vertical whitespace please.

src/system.c
```diff
@@ -123,7 +123,26 @@ void sys_set_seccomp_syscall(bool enable)
...
+
+	nr_seccomp = arch_syscall_resolve_name(arch_def_native, "seccomp");
+	if (nr_seccomp != __NR_SCMP_ERROR) {
```
pcmoore: Instead of directly querying to see if seccomp(2) is supported, we should use the sys_chk_seccomp_syscall() function so we honor any configured API level settings.

While we are on the topic, I'm wondering if the KILL_PROCESS capability should be optionally gated behind an API level, thoughts?  (Full disclosure: I haven't thought much about it yet, just wanted to mention the idea.)

src/system.c
pcmoore: I would look at what we did for SCMP_ACT_LOG and duplicate that, I believe it should do everything you need.

src/system.c
pcmoore: Also ... and I should have mentioned this first, my apologies ... look at what we did to support SCMP_ACT_LOG.

pcmoore: Regarding "API levels", if KILL_PROCESS went in at the same time as ACT_LOG, we may be able to just reuse the existing API level ("3", see include/seccomp.h.in).

src/system.c
pcmoore: Nevermind, I see you define it in the patch below (src/system.h).

pcmoore: Actually no, scratch that ... we could still be on a non cutting edge system which has seccomp.h but doesn't have SECCOMP_RET_ACTION_FULL defined; see how we handle SECCOMP_RET_LOG and friends in src/system.h.

tests/Makefile.am
```diff
@@ -168,7 +170,8 @@ EXTRA_DIST_TESTCFGS = \
 	39-basic-api_level.tests \
 	40-sim-log.tests \
 	41-sim-syscall_priority_arch.tests \
-	42-sim-adv_chains.tests
+	42-sim-adv_chains.tests \
+	43-sim-kill_process.tests
 
 EXTRA_DIST_TESTSCRIPTS = \
 	38-basic-pfc_coverage.sh 38-basic-pfc_coverage.pfc
```
pcmoore: What would it take to add a "live" test for the KILL_PROCESS action?

src/system.c
```diff
@@ -124,8 +124,10 @@ void sys_set_seccomp_syscall(bool enable)
 int sys_chk_seccomp_action(uint32_t action)
 {
 	int rc, nr_seccomp;
-	uint32_t kaction = action & SECCOMP_RET_ACTION_FULL;
 
+/* SECCOMP_RET_ACTION_FULL was defined in kernel v4.14 */
+#ifdef SECCOMP_RET_ACTION_FULL
+	uint32_t kaction = action & SECCOMP_RET_ACTION_FULL;
```
pcmoore: See my earlier comments about SECCOMP_RET_LOG in src/system.h.

src/gen_pfc.c
```diff
@@ -110,6 +110,10 @@ static void _pfc_arg(FILE *fds,
 		fprintf(fds, "$a%d", node->arg);
 }
 
+#ifndef SECCOMP_RET_ACTION_FULL
+#define SECCOMP_RET_ACTION_FULL 0xffff0000U
+#endif
+
```
pcmoore: Let's just `#include "system.h"` and not duplicate the #define.

pcmoore:
Thanks for the code @drakenclimber!

I just made a quick pass through the patches, let me know if you have any questions/comments on my feedback.

drakenclimber:
Thanks, @pcmoore.  Your comments are spot on.  Thanks.

I'll get to work on making the requested changes first thing tomorrow.

pcmoore:
A quick note to let you know I haven't forgotten about this, I'm just a bit busy with SELinux and audit kernel patch review at the moment.  Things should calm down a bit in the next week or two as the queue gets flushed.

Thanks again for all the help and patches!

drakenclimber:
No worries.  In the meantime, I'll try to resolve the A2 ordering bug I uncovered.

drakenclimber:
I'm closing this pull request and opening a new one based on the latest branch

PR #115, opened 2018-03-23
Issues/112

drakenclimber:


drakenclimber:
The above code coverage failures are because I can't make the new db_chain_lt() - either in macro or function form - return false.  This is a little disconcerting to me.

drakenclimber:
> The code coverage failures are because I can't make the new db_chain_lt() - either in macro or function form - return false.  This is a little disconcerting to me.

Disregard the above.  I was trying to multitask yesterday... and failed miserably at it.  I made a silly mistake when I ported db_chain_lt() from a macro to a function.  After fixing that and adding a test designed to exercise db_chain_lt() and _db_tree_add(), all is looking good.

Phew.

pcmoore:
Thanks so much for this @drakenclimber!  You get extra bonus points for increasing the code coverage too.

I just took a quick look at the final changes and they all look reasonable, although the way they are split across the different patches is a little unusual for libseccomp.  I suspect I can just squash some of them together to make things a bit more sane, I'd like to avoid giving you more work at this point :)

Updates later today ...

pcmoore:
Still working, but I merged the db chain ordering fix in cf98f79d0894221beb9f2753c092304237617c1c and the new `43-*` and `44-*` tests in 0f03fdf63d5a33ed1a9b0e9e8dbdd4311aa3c48f.  You will note that I squashed a number of @drakenclimber's commits into these two patches (basically squash the fixes in with the base commit).

drakenclimber:
> I just took a quick look at the final changes and they all look reasonable, although the way they are split across the different patches is a little unusual for libseccomp.

Yeah, sorry about that.  The fix definitely evolved over time.  It's not how I would have sent it to the kernel guys upstream, but github allows for a little more creativity.  :) Thanks for being flexible.

pcmoore:
Okay, I think 475f67ccd5814d94c935e76d3aeb743ce257f2d3 is the last commit I want to merge from this PR.  As we talked about previously, I think a lot of the macros in src/db.h are long overdue for conversion into proper functions, but I'd like to do them all and once, and investigate if they even need to be in src/db.h (e.g. is src/db.c the only user?).

Is that okay with you @drakenclimber ?

pcmoore:
> Yeah, sorry about that. The fix definitely evolved over time. It's not how I would have sent it to the kernel guys upstream, but github allows for a little more creativity. :) Thanks for being flexible.

When I make people wait this long to get their PR's accepted I figure I should bear some of the burden. 
 Not to mention that you took the responsibility for turning a simple gist of mine into a properly tested patch ;)

pcmoore:
I fixed up the `db_chain_*` macros in commit b241edec6d81da56016b0f48341c15b50c11ecb5 so I'm going to consider this closed.  Thanks for all your help @drakenclimber; if you noticed I've missed anything from the PR please go ahead and reopen it.

PR #121, opened 2018-06-02
BUG: remove static linking of tools

ffontaine:
Do not force static link of tools, it breaks build with:
BR2_SHARED_LIBS=y

Patch retrieved from
https://git.buildroot.net/buildroot/tree/package/libseccomp/0001-remove-static.patch and slighly updated to work with 2.3.3

Signed-off-by: Bernd Kuhls <bernd.kuhls@t-online.de>
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/tools/Makefile.am
+++ b/tools/Makefile.am
@@ -35,8 +35,6 @@ scmp_bpf_disasm_SOURCES = scmp_bpf_disasm.c bpf.h util.h
 scmp_bpf_sim_SOURCES = scmp_bpf_sim.c bpf.h util.h
 
 scmp_sys_resolver_LDADD = ../src/libseccomp.la
-scmp_sys_resolver_LDFLAGS = -static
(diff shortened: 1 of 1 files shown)


pcmoore:
Hi @ffontaine, I apologize it has taken so long to comment on your PR.  This is likely a carryover from the pre-autotools build and I wanted to make sure we were using the library from the current build and not the system's installed library; I *believe* autotools takes care of this for us now and we can remove this.

@drakenclimber any comments and/or objections?

drakenclimber:
Looks good to me.  I tried to break this and make it fall back to using the installed library, but I couldn't do it.  I'm no autotools expert, but I think @pcmoore is correct and autotools is doing the right thing.

pcmoore:
Merged via 1052ec1cc15be55f679f0d93496226f247d179b4, thanks again.

FYI: I also removed the '-static' flag from the `scmp_api_level` build.

PR #124, opened 2018-06-22
RFE: refactor bpf duplicate sorting code

drakenclimber:
In _gen_bpf_arch(), there was an identical block of code to sort
the primary database syscalls and the secondary database
syscalls.  This commit refactors those duplicated, inline loops
into a single function.

### DIFF
--- a/src/gen_bpf.c
+++ b/src/gen_bpf.c
@@ -1209,6 +1209,71 @@ static struct bpf_blk *_gen_bpf_syscall(struct bpf_state *state,
 }
 
 /**
+ * Sort the syscalls
+ * @param syscalls the linked list of syscalls to be sorted
+ * @param s_head the head of the linked list to be returned to the caller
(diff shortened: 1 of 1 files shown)


drakenclimber:
I forgot all about this pull request.  It's not critical by any means, but it does simplify things a bit.  If done correctly, it should have no change on generated filters.

pcmoore:
Full disclosure: I forgot about it too - sorry.

I think it's a reasonable change, but I see some overlap with the binary tree PR you are working on; do you want me to review/merge this anyway, or hold off?

drakenclimber:
I honestly don't have a strong preference either way.

When I submitted this, I thought I may have several refactors/changes that I would need to make before adding the binary tree code.  This ended up being the only refactor of any significance that I found.

I would say let's close it for now and see if we can pick it up with the binary tree changes.

pcmoore:
Okay, sounds good.

PR #128, opened 2018-09-08
x32: fix broken tests

1camper:
Many tests are broken on x32 because the used syscalls are given as a number and - without X32_SYSCALL_BIT set - don't pass the arch filter, e.g.

Test 01-sim-allow%%001-00001 result: FAILURE bpf_sim resulted in KILL
Test 01-sim-allow%%001-00002 result: FAILURE bpf_sim resulted in KILL
Test 01-sim-allow%%001-00003 result: FAILURE bpf_sim resulted in KILL
...
Regression Test Summary
tests run: 5077
tests skipped: 101
tests passed: 3374
tests failed: 1703
tests errored: 0

### DIFF
--- a/src/arch-x32.c
+++ b/src/arch-x32.c
@@ -30,7 +30,7 @@ const struct arch_def arch_def_x32 = {
 	.token = SCMP_ARCH_X32,
 	/* NOTE: this seems odd but the kernel treats x32 like x86_64 here */
 	.token_bpf = AUDIT_ARCH_X86_64,
-	.size = ARCH_SIZE_32,
+	.size = ARCH_SIZE_64,
 	.endian = ARCH_ENDIAN_LITTLE,
 	.syscall_resolve_name = x32_syscall_resolve_name,
 	.syscall_resolve_num = x32_syscall_resolve_num,
--- a/tests/02-sim-basic.tests
+++ b/tests/02-sim-basic.tests
@@ -18,6 +18,8 @@ test type: bpf-sim
 02-sim-basic	x86	174-350		N		N		N	N	N	N	KILL
 02-sim-basic	x86_64	4-14		N		N		N	N	N	N	KILL
 02-sim-basic	x86_64	16-350		N		N		N	N	N	N	KILL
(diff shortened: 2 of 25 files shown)


pcmoore:
There are a things in this PR that make me nervous, the most significant being that I don't believe we should need to add the x32 bit to the syscalls used in the tests, especially the tests run only in the simulator.

Unfortunately it has been a while since I've tested directly on x32 (I lost my x32 testing VM in a disk crash), can you provide some more information on how you tested this?  What distro are you using that provides x32 support?

1camper:
The filters used for the simulator are the same as the ones used when running live. So the same restrictions apply.

Tested with Gentoo using a recent x32 stage3 tarball for a chroot environment (or install fully following the handbook).

pcmoore:
> The filters used for the simulator are the same as the ones used when running live. So the same restrictions apply.

No, the simulated and live tests are different.  Regardless, we didn't need to explicitly add the x32 bit in the past so we shouldn't now; there must have been some regression in the code.

> Tested with Gentoo using a recent x32 stage3 tarball for a chroot environment (or install fully following the handbook).

Okay, thanks.  That is what I was using, I was just hoping there was a more plug-n-play option available, although I suspect x32 has probably gotten even more niche that it was when it was first introduced.

pcmoore:
I apologize it too me so long to get back to this @1camper, I needed to recreate my Gentoo x32 install.

I'm now seeing the failures you are seeing, and yes the automated tests are pretty badly broken on x32.  As you pointed out, many (all?) of the failures stem from the fact that a large number of the tests do not use x32 syscalls (the X32 bit is not set) and are caught in the libseccomp BPF preamble where we do the ABI checks.  At this point I'm not sure of the "proper" fix, but my gut feeling is that it should be limited to the tests/*.tests files; in other words, just disable the offending tests on x32.

pcmoore:
I've created issue #141 to track this problem.

pcmoore:
I'm going to close this PR because I don't believe this is the proper way to fix this problem moving forward.  At the very least the tests should now run clean on x32.

@1camper if you are still having problems, please let us know.

Issue #117, opened 2018-05-12
BUG: seccomp_export_pfc hangs on prioritized syscalls with no rules


varqox:
The program below hangs on seccomp_export_pfc() if the priority of a syscall was set, but no rule has been added for it.
```c
#include <stdio.h>
#include <seccomp.h>

int main() {
	scmp_filter_ctx ctx = seccomp_init(SCMP_ACT_KILL);
	if (ctx == NULL) {
		perror("seccomp_init()");
		return 1;
	}

	// Set the x86 architecture
	if (seccomp_arch_native() != SCMP_ARCH_X86) {
		seccomp_arch_add(ctx, SCMP_ARCH_X86);
		seccomp_arch_remove(ctx, SCMP_ARCH_NATIVE);
	}

	int rc = seccomp_syscall_priority(ctx, SCMP_SYS(mmap2), 255);
	// seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(mmap2), 0);

	int stdout_fileno = 1;
	seccomp_export_pfc(ctx, stdout_fileno); // Hangs here

	return 0;
}
```

drakenclimber:
Looks like the "if" statement at line [307 of gen_pfc.c](https://github.com/seccomp/libseccomp/blob/master/src/gen_pfc.c#L307) is the culprit.  The syscall is invalid, so we "continue", but we don't advance p_iter.

```
	while (p_iter != NULL) {
		if (!p_iter->sys->valid)
			continue;
		_gen_pfc_syscall(db->arch, p_iter->sys, fds);
		p_iter = p_iter->next;
	}
```

pcmoore:
Thanks for the report @varqox, and I believe @drakenclimber is right about the root cause.  I'll queue this up for later in the week, but if someone wants to send me a PR all the better ;)

drakenclimber:
@pcmoore - I'll try to get to this in the next couple days

pcmoore:
Thanks @drakenclimber!  No worries if you can't get to it, just let me know.

pcmoore:
Fixed with PR #131.

PR #131, opened 2018-09-19
pfc: fix PFC export hang on prioritized syscall with no rules (GH issue #117)

drakenclimber:
github user @varqox reported that generating PFC will hang if the
libseccomp filter contains a syscalle with a priority but no rule
set.  The root cause is the while() loop in gen_pfc.c that walks
through the filter's syscalls.  It  wasn't properly advancing
through the list when p_iter was invalid.

Addresses issue #117 

### DIFF
--- a/tests/38-basic-pfc_coverage.c
+++ b/tests/38-basic-pfc_coverage.c
@@ -81,6 +81,12 @@ int main(int argc, char *argv[])
 	if (rc < 0)
 		goto out;
 
+	/* bug #117 - seccomp_export_pfc hangs on prioritized syscalls
+	 * with no rules */
+	rc = seccomp_syscall_priority(ctx, SCMP_SYS(poll), 255);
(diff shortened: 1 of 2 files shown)


tests/38-basic-pfc_coverage.c
```diff
@@ -81,6 +81,12 @@ int main(int argc, char *argv[])
 	if (rc < 0)
 		goto out;
 
+	/* bug #117 - seccomp_export_pfc hangs on prioritized syscalls
+	 * with no rules */
```
pcmoore: I try not to reference GitHub specific numbers since those really only have meaning in GitHub and who knows where this code might be 5, 10, 15 years from now.

No need to respin, just something to keep in mind for future patches.

pcmoore:
Merged via 6646e21ed2734dca355c5b550cb45f0379330e02, thanks!

Issue #86, opened 2017-06-20
BUG: seccomp_arch_add() returns -EEXISTS on endian mismatch


mvo5:
(note that my issue description starts with golang but its actually a C issue, see below)

I was writing some unit tests today that exercise ScmpFilter.AddArch(seccomp.ArchPPC) on my amd64 system. This did not return any error, however the architecture was not added to the filter (as visible via exportPFC).

Here is a trivial reproducer (needs to run on amd64 or i386):
```
package main

import (
	"os"
	"github.com/seccomp/libseccomp-golang"
)

func main() {
	secFilter, err := seccomp.NewFilter(seccomp.ActKill)
	if err != nil {
		panic(err)
	}
	err = secFilter.AddArch(seccomp.ArchPPC)
	if err != nil {
		panic(err)
	}
	secFilter.ExportPFC(os.Stdout)
}
```

After a bit of debugging it turned out that seccomp_arch_add() will return EEXIST if there is an endian mismatch. In db.c:db_col_db_add() there is:
```
if (col->endian != 0 && col->endian != db->arch->endian)
		return -EFAULT;
```

The golang code (rightfully) ignores EEXIST which leads to the behaviour I have observed.

I wonder if it would make sense that seccomp_arch_add() would return a different error code, maybe EINVAL or something? If this is too dangerous (as it might break existing applications), maybe it could be documented better. I am happy to provide a PR.

pcmoore:
Thanks for reporting this, I'll have to take a closer look.

pcmoore:
I'm very sorry it has taken me so long to get back to this @mvo5, I appreciate your patience.

To clarify things a bit, the relevant current db.c:db_col_db_add() looks like this:
```
        if (col->endian != 0 && col->endian != db->arch->endian)
                return -EEXIST;
```
... I think the version above in the original problem report was a locally patched copy to work around the problem.  That said, changing the returned error here does sound reasonable; we've been using EDOM in at least some of the other arch/endian code, does that sound reasonable to you?

What do you think @mheon?

It looks like we should probably also update db.c:db_col_merge() while we are at it.

mheon:
From an API standpoint, I definitely think it makes good sense to make the change - overloading error codes is always problematic for debugging.

On the libseccomp-golang side, this shouldn't require any code changes, so long as the negative ERRNO convention is maintained; maybe a few lines of comments on AddArch to explain what the new error means, so the API documentation will be complete.

drakenclimber:
[...]

PR #138, opened 2019-02-18
db: Return -EDOM on endian mismatch during arch add

drakenclimber:
This commit clarifies the error code when seccomp_arch_add() or
seccomp_merge() fails due to an endian mismatch.  Previously,
libseccomp would return -EEXIST if the new architecture's
endianness did not match.

This addresses GitHub Issue #86 - BUG: seccomp_arch_add() returns
-EEXISTS on endian mismatch

Reported-by: Michael Vogt <michael.vogt@gmail.com>
Suggested-by: Paul Moore <paul@paul-moore.com>


pcmoore:
Merged via 4a35b6ea6f7c836734536420c50a2745a9e24c69, thanks.

PR #144, opened 2019-03-05
RFE: header updates

tych0:
There's a bunch of stuff that has been added in recent kernels, let's add it here.

### DIFF
--- a/src/system.h
+++ b/src/system.h
@@ -117,6 +117,9 @@ typedef struct sock_filter bpf_instr_raw;
 #ifndef SECCOMP_FILTER_FLAG_LOG
 #define SECCOMP_FILTER_FLAG_LOG			(1UL << 1)
 #endif
+#ifndef SECCOMP_FILTER_FLAG_SPEC_ALLOW
+#define SECCOMP_FILTER_FLAG_SPEC_ALLOW		(1UL << 2)
+#endif
 
 #ifndef SECCOMP_RET_LOG
 #define SECCOMP_RET_LOG		0x7ffc0000U /* allow after logging */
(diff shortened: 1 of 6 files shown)


tych0:
Oh, but I guess we should define them in include/seccomp.h.in too, let me do that.

src/system.h
```diff
@@ -61,6 +61,7 @@ struct db_filter_col;
 #define SECCOMP_RET_TRAP	0x00030000U /* disallow and force a SIGSYS */
 #define SECCOMP_RET_ERRNO	0x00050000U /* returns an errno */
 #define SECCOMP_RET_TRACE	0x7ff00000U /* pass to a tracer or disallow */
+#define SECCOMP_RET_LOG		0x7ffc0000U /* allow after logging */
```
pcmoore: We added SECCOMP_RET_LOG back in 3b22b153c70ef00044fb309d9a9fecc12b777e98, does that commit not work for you?

pcmoore: I suppose we could add this bit of the patch, but drop the section below as commit 3b22b153 should have that covered.

pcmoore:
Thanks.

I merged the first two patches via 635496a5938a57bd0b3877f59504b7e7d07b37c0 and b09e6e800646de6f5cbee4efeb02c5b1ee4e09cf.  I'm going to leave the SPEC_ALLOW changes for after the v2.4 release.

src/system.h
tych0: This duality between system.h and seccomp.h.in is a little bit overengineered, IMO.

However, I'm going to import SECCOMP_RET_USER_NOTIF into system.h (for testing that the return type is supported) in the user-notification series, so I think it makes sense to add this one here as well, to keep things in sync with the kernel headers. Although yes, it's a no-op for this series.

pcmoore:
Since the SPEC_ALLOW stuff is part of the user notification PR let's close this out.  @tych0 if you feel otherwise feel free to comment and we can reopen this PR.

PR #149, opened 2019-04-10
Issues/148v2

drakenclimber:


tests/50-sim-hash_collision.c
```diff
@@ -0,0 +1,91 @@
...
+		goto out;
+
+	/* These rules were derived from an issue reported by Tor:
+	 * https://github.com/seccomp/libseccomp/issues/148
+	 *
+	 * In the steps below, syscall 1001 is configured similarly to how
+	 * Tor configured socket.  The fairly complex rules below led to
+	 * a hash collision with rt_sigaction (syscall 1000) in this test.
+	 */
+
```
pcmoore: I think we should mention the current hash implementation (MurmurHash3) both in the commit message as well as in the tests themselves (both .c and .py) as this test is going to be dependent on the hashing function.

pcmoore:
GitHub doesn't seem to provide a mechanism to comment on patch descriptions so I'll leave this here: in the test patch, I would suggest providing a full URL instead of simply "issue 148" as the URL provides some context for the issue number.

pcmoore:
Somewhat similar comment regarding the issue number in the first patch as well, although there you at least mention GitHub so it isn't as critical.

pcmoore:
This is nitpicky too, but I prefer the commit description I provided in my original patch:

> This addresses a problem where dissimilar instruction blocks were improperly hashed to the same value because we were not taking into account the accumulator state.

I worry your description is overly specific about the particular bug we are seeing with the current Tor filter and doesn't correctly describe the problem in general (ignoring the accumulator state).

pcmoore:
My apologies if these comments seem a bit minor, but now that you're a maintainer, the bar for acceptance gets raised a bit higher ;)

drakenclimber:
No worries.  I appreciate the feedback.  I admit that I focused less on the commit messages and primarily on the content - especially finding a test case that reproduces their issue.

I'll rework the patch with your comments addressed.  Thanks!

Issue #148, opened 2019-03-24
BUG: Tor sandbox issue with libseccomp v2.4.0


toralf:
There's a small regression with 2.40 and Tor:

https://trac.torproject.org/projects/tor/ticket/29819#comment:5

and I do wonder if this is a shortcoming of the Tor code?

pcmoore:
Hi @toralf, unfortunately I'm not very familiar with the Tor sandboxing code.  Is there anyway for you to capture the libseccomp PFC output from the Tor libseccomp filter so we can compare it with the desired Tor filter?

See the `seccomp_export_pfc()` function as a way of capturing the PFC output.

pcmoore:
@drakenclimber you wouldn't happen to have any experience with Tor, would you?

toralf:
> Hi @toralf, unfortunately I'm not very familiar with the Tor sandboxing code
Well, seems to be an issue tih the Tor code, can be reproduced there now at different systems too.
Thx for your help.

pcmoore:
Thanks for keeping us updated @toralf, if you need any help/advice on writing libseccomp filters for Tor (or any other application) please let us know, we're happy to help.

Good luck!

toralf:
I do wonder about your opinion about https://trac.torproject.org/projects/tor/ticket/29819#comment:10 ?

pcmoore:
Hello again.  As I said previously, I'm really not very familiar with the Tor libseccomp filters and there isn't much information in the linked ticket.  I think it might be helpful if you could capture the libseccomp PFC output from the Tor libseccomp filters so we can compare it to the desired filter.

pgerber:
I got the PFC output ([pfc.txt](https://github.com/seccomp/libseccomp/files/3024632/seccomp.txt)).

rt_sigaction(1 (=SIGHUP), …) is what causes the crash but I'm not sure if the PFC is trying to tell me that that's allowed or not:

```
  # filter for syscall "rt_sigaction" (13) [priority: 65526]
  if ($syscall == 13)
    if ($a0.hi32 == 0)
      if ($a0.lo32 == 25)
        action ALLOW;
      if ($a0.lo32 == 17)
        action ALLOW;
      if ($a0.lo32 == 15)
        action ALLOW;
      if ($a0.lo32 == 13)
        action ALLOW;
      if ($a0.lo32 == 12)
        action ALLOW;
      if ($a0.lo32 == 10)
        action ALLOW;
      if ($a0.lo32 == 2)
[...]

PR #150, opened 2019-04-11
BUG: issues/148v3

drakenclimber:
Changes from v2 to v3:

- Addressed code review comments from @pcmoore
- Fully incorporated the _hsh_add() patch proposed by @pcmoore in issue #148.  (I hadn't updated from an earlier iteration.)
- Added 50-sim-hash_collision to tests/.gitignore

### DIFF
--- a/src/gen_bpf.c
+++ b/src/gen_bpf.c
@@ -141,6 +141,9 @@ struct bpf_blk {
 
 struct bpf_hash_bkt {
 	struct bpf_blk *blk;
+	struct acc_state acc_start;
+	struct acc_state acc_end;
+
 	struct bpf_hash_bkt *next;
 	unsigned int found;
 };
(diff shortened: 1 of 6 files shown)


src/gen_bpf.c
```diff
@@ -569,11 +572,16 @@ static int _hsh_add(struct bpf_state *state, struct bpf_blk **blk_p,
...
+	h_val_tmp[0] = hash(blk->blks, _BLK_MSZE(blk));
+	h_val_tmp[1] = hash(&blk->acc_start, sizeof(blk->acc_start));
+	h_val_tmp[2] = hash(&blk->acc_end, sizeof(blk->acc_end));
+	h_val = hash(h_val_tmp, sizeof(h_val_tmp));
 	blk->hash = h_val;
 	blk->flag_hash = true;
 	blk->node = NULL;
 	h_new->blk = blk;
+	h_new->acc_start = blk->acc_start;
+	h_new->acc_end = blk->acc_end;
```
pcmoore: I actually liked what you were doing in your patch where you simply reused the accumulator state kept inside the `blk` variable.  I'd vote for changing this to use that approach, or was there a problem with that?

pcmoore:
Other than the small nit above, this looks good to me - thanks @drakenclimber.

src/gen_bpf.c
drakenclimber: I don't believe there were any issues with the approach I had taken, but I admit I would like to get more testing miles on it.

I had forgotten to push your latest patch to my git repo and was simply running with my intermediate one.  I'll grab the reused accumulator changes from my patch and merge them with this one.  I'll try to log some more testing miles as well.

Stay tuned...

pcmoore: Great, thanks!

PR #154, opened 2019-05-02
BUG: define the Cython "language_level"

pcmoore:
Set the Cython language level to "3str" which is described in the
Cython 0.29 changelog:

 "A new language level name 3str was added that mostly corresponds to
  language level 3, but keeps unprefixed string literals as type ‘str’
  in both Py2 and Py3, and the builtin ‘str’ type unchanged. This will
  become the default in the next Cython release and is meant to help
  user code a) transition more easily to this new default and
  b) migrate to Python 3 source code semantics without making support
  for Python 2.x difficult."

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -109,8 +109,8 @@ AC_ARG_ENABLE([python],
 	[build the python bindings, requires cython])])
 AS_IF([test "$enable_python" = yes], [
 	# cython version check
-	AS_IF([test "$CYTHON_VER_MAJ" -eq 0 -a "$CYTHON_VER_MIN" -lt 16], [
-		AC_MSG_ERROR([python bindings require cython 0.16 or higher])
+	AS_IF([test "$CYTHON_VER_MAJ" -eq 0 -a "$CYTHON_VER_MIN" -lt 29], [
+		AC_MSG_ERROR([python bindings require cython 0.29 or higher])
 	])
 	AM_PATH_PYTHON
 ])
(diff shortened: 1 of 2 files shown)


pcmoore:
@drakenclimber can you take a look at this PR, it fixes #154.

drakenclimber:
Will do.

drakenclimber:
Oh, and I reviewed the cython versions and documentation regarding 3str.  Looks like you nailed it.

pcmoore:
@drakenclimber thanks for the quick review.

From what I can see, it looks like "3str" should be safe for both Python 3.x and 2.7 so I don't think we will have a real problem there (everyone should be on those Python versions, or greater, by now).  However, I am a little concerned about requiring Cython v0.29; I figure we can mitigate this somewhat by not backporting this to the release-2.4 branch.  Thoughts on this @drakenclimber?

drakenclimber:
> However, I am a little concerned about requiring Cython v0.29

FWIW, I often run a stable version of Oracle Linux (i.e. not necessarily the cutting edge), and I had easy access to Cython v0.29 via `yum install pip` and then `pip install Cython`.  If I have relatively easy access to it, I would think most users also would.

> I figure we can mitigate this somewhat by not backporting this to the release-2.4 branch. Thoughts on this @drakenclimber?

Sounds like a good plan to me.  It's something we should fix, but waiting for v2.5 is a good compromise.

pcmoore:
Okay, looks like we're on the same paged.  Merged via d390edad9a8540c2e2dd0b12732cc8dd3fe1cc69.

Issue #153, opened 2019-05-01
RFE: seccomp_rule_add became very slow since v2.4.0


varqox:
Hello,
the issue was introduced by the commit ce3dda9a1747cc6a4c044eafe5a2eb653c974919 between v2.3.3 and v2.4.0. Consider the following example: [foo.c.zip](https://github.com/seccomp/libseccomp/files/3135012/foo.c.zip).
It adds a very large number of rules. And works around 100 times slower after the abovementioned commit.

`foo.c` execution time using v2.4.1: `0.448`
`foo.c` execution time using v2.3.3: `0.077`

I dug a little and found out that db_col_transaction_start() copies already existing filter collection and uses arch_filter_rule_add() to duplicate filter rules. But arch_filter_rule_add() calls arch_syscall_translate() which calls arch_syscall_resolve_name() which works in O(number of syscalls on the given architecture). So adding one rule works at least in O(number of already added rules * number of syscalls on used architectures) which IMO is really bad.
I counted the number of calls to arch_filter_rule_add() in the above example and it equals `201152`.

Before that commit number of calls to arch_filter_rule_add() was `896`. And from what I understand from the code, also db_col_transaction_start() copies already existing filter collection and does not use arch_filter_rule_add(). Which gives us estimation: time of adding a rule around O(number of already added rules + number of syscalls on the given architecture), which is much better.

However IMO it should not be related to the number of already added rules, because adding n rules works in O(n^2) then. But that is a topic to a different discussion, so it should not be a problem for small filters or filters generated infrequently.

Why this problem matters?
Some filters need executing programs PID (e.g. allowing the thread to send signals only to itself). So if the restricted program needs to be executed a considerable number of times, it becomes a very visible overhead. I have a filter of around 300 rules and libseccomp overhead is around 0.16s per run of the sandboxed process (I run the process dozens of times).

Thank you in advance for your help!

pcmoore:
Hi @varqox.

Yes, the syscall resolver functions could use some improvement, in fact if you look at the code you will see several comments like the following:
```
/* XXX - plenty of room for future improvement here */
```
If you want to look at improving that code, we should could use the help!

drakenclimber:
As @pcmoore mentioned there are ample opportunities for speeding up the _creation_ of a seccomp filter using libseccomp.  Your research above outlined one of several areas that could use improvement.  This has not been a concern for my users, so I have not focused on it.

[...]

PR #156, opened 2019-05-02
RFE: speed up resolving syscalls by name by using hash map

varqox:
It slightly mitigates #153.

### DIFF
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -1,3 +1,5 @@
 libseccomp.a
 arch-syscall-check
 arch-syscall-dump
+arch-*-syscall-hashmap-printer
+arch-*-syscall-hashmap.c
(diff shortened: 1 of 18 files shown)


pcmoore:
I'm not entirely sure how I feel about this approach at the moment.

If I understand you correctly @varqox, the main motivation for this was issue #153, which I believe we've addressed in the master branch (scheduled for the v2.5 release).  How do you feel about this PR now?

varqox:
Yes, I did this because before the fix from #180 resolving the name of a syscall was the primary thing that libseccomp did (based on `perf` analysis). Now it is only roughly 10% of what libseccomp is doing.
Moreover, my solution reduces name resolving only by about 50%. Also, it is not a clean implementation because of the lack of compile-time data manipulation (building the hash map) in C.

For now, I don't think that this solution is worth ~5% performance increase of libseccomp.

pcmoore:
> For now, I don't think that this solution is worth ~5% performance increase of libseccomp.

That is my thinking at the moment too.  While there are definitely improvements to be had with respect to syscall name/number resolution, I'm not sure this is the right fix (although a hash map is likely a good idea).

With the conversation above in mind, I'm going to close this PR.  Thanks again for your patience and help in resolving the #153.

PR #157, opened 2019-06-22
RFE: add support for io-uring related system calls in kernel 5.1

omegacoleman:
This patch adds support for the newly-introduced io-uring system calls
in kernel 5.1, for description with io-uring, see:
https://lwn.net/ml/linux-fsdevel/20190112213011.1439-1-axboe@kernel.dk/

I was doing this because this enables us to configure and enable it in docker (opencontainer/runc in paticular) without disabling seccomp.

Signed-off-by: Stephen Coleman <omegacoleman@gmail.com>


pcmoore:
This looks okay to me, if it looks good to @drakenclimber too I think we can get this merged.

drakenclimber:
Looks good to me.  I compared the values by hand and also ran the patch through arch-syscall-validate.

pcmoore:
Merged via 5fc22428507ecea00ee9e2215d972777da9a99b6, thanks!

omegacoleman:
good to hear

PR #159, opened 2019-07-15
BUG: install the python extension to the root package dir

debfx:
Commit 8ad3638ea9023c3948976dfadebd1554380a31c9 effectively added libseccomp/
to the install path of the python extension.
This changed the import module name from "seccomp" to "libseccomp.seccomp",
breaking existing users.

Revert the install path like it was before 2.4.0


pcmoore:
Sorry for the delay; thanks for reporting this and providing a fix!  It looks good to me so I've merged it via ff7fc73bdf4602323ce320a2094c006c9c3e4d42.

Issue #27, opened 2016-02-12
RFE: consider removing negative (non-canonical) __NR_xyz defines


amluto:
IMO __NR_xyz for syscall xyz should either match the actual kernel-assigned number or should not be defined.  This code, for example, can cause confusion:

```
#define __PNR_userfaultfd   -10200
#ifndef __NR_userfaultfd
#define __NR_userfaultfd    __PNR_userfaultfd
#endif /* __NR_userfaultfd */
```

libseccomp is not the sole user of __NR_xyz defines.

Using LIBSECCOMP_xyz or sticking with __PNR_syz would be fine.

amluto:
See https://github.com/sandstorm-io/sandstorm/pull/1495#issuecomment-182392731

pcmoore:
@amluto Out of curiosity, in the sandstorm issue you mentioned the following:

> At some point we should drop libseccomp.  It's problematic for a few reasons.

Other than the issue you mention here, can you explain the others?  We would like to make libseccomp useful for other projects, and if it isn't it would be good to know why.

amluto:
IIRC the main issue I had was that filtering on anything other than syscall nr is very inflexible.  For example, I couldn't express:

Block init_module.
Allow socket, but only for the following address families

unless I inverted it and disallowed families above the largest family number I wanted and also disallowed families in between the ones that are okay.

kentonv:
@dwrensha is reporting that if he builds a binary on a system whose headers are old and don't define the __NR_blah constant for a syscall, and so instead gets libseccomp's pseudo-number, the result is that the seccomp filter *does not filter the syscall*. Does this seem correct? If so this seems like a serious security problem.

dwrensha:
Here is an example program:

```C
#include <stdio.h>
#include <unistd.h>
#include <errno.h>
#include <seccomp.h>
#include <string.h>
#include <fcntl.h>

int main() {
  scmp_filter_ctx ctx = seccomp_init(SCMP_ACT_ALLOW);
  if (ctx == NULL) return 1;

  int result = seccomp_rule_add(ctx, SCMP_ACT_ERRNO(EACCES), SCMP_SYS(userfaultfd), 0);
  if (result < 0) return 1;

  result = seccomp_load(ctx);
  if (result < 0) return 1;

  long result2 = syscall(323, O_NONBLOCK);  // 323 = userfaultfd() on x86_64
  if (result2 >= 0) {
[...]

PR #172, opened 2019-10-01
api: stop defining __NR_x values for syscalls that don't exist

pcmoore:
Historically libseccomp has created a __NR_x definition for every
syscall it supports, even those that aren't valid for a given ABI.
While this seemed like a good idea at the time, it turned out to have
some unwanted and nasty side effects.  This patch finally corrects
this problem.

The basic approach is quite simple: move the SCMP_SYS() macro to use
__SNR_x values instead of __NR_x values.  The unfortunate side effect
of this change is that instead of just worrying about #defines for the
__PNR_x values we now have to have a __SNR_x define for *every*
syscall.  The good news is that after this patch that should only be
a few new syscalls every year - a very manageable task.


pcmoore:
This should resolve issue https://github.com/seccomp/libseccomp/issues/27.

drakenclimber:
Looks good to me.  The move away from __NR_xyz was much cleaner than I expected.  Nice work

pcmoore:
Thanks for taking a look, merged via bf162816a6e42d7a40e54aadc18d1a98c34ca971.

PR #186, opened 2019-11-13
BUG: api: define __SNR_ppoll again

mlichvar:
It seems commit bf747eb21e428c2b3ead6ebcca27951b681963a0 accidentally removed the
__SNR_ppoll definition, which breaks building of applications using SCMP_SYS(ppoll). The patch just adds the definition back.

### DIFF
--- a/include/seccomp-syscalls.h
+++ b/include/seccomp-syscalls.h
@@ -272,6 +272,7 @@
 #define __PNR_timerfd_gettime64			-10238
 #define __PNR_timerfd_settime64			-10239
 #define __PNR_utimensat_time64			-10240
+#define __PNR_ppoll				-10241
 
 /*
  * libseccomp syscall definitions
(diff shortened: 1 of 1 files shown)


pcmoore:
Hi @mlichvar, thanks for catching this and submitting a PR - I'm sorry we broke this in the last update :(

Unfortunately, I don't think this patch is quite right, I think we need to do something like the following:

```
#define __PNR_ppoll			-10241
```

... and this:

```
#ifdef __NR_ppoll
#define __SNR_ppoll			__NR_ppoll
#else
#define __SNR_ppoll			__PNR_ppoll
#endif
```

Would you be willing to respin/test your PR with the above changes?

mlichvar:
Ok, I've made the suggested change. I assume this is needed because the definition in the kernel headers is now wrapped in __ARCH_WANT_TIME32_SYSCALLS. Thanks.

pcmoore:
Merged via e3647f5b6b52996bf30d0c2c1d1248e4182e1c1c, thanks!

cpaelzer:
FYI - tested on top of 2.4.2 and working find with chrony.
Thanks @mlichvar to submit that after our discussion.

Just so you know, this issue also will trigger a bunch of systemd tests to fail, but fortunately the fix also resolves them.

PR #182, opened 2019-11-06
BUG: fix various bugs found when working on code coverage

pcmoore:
While working to claw back some code coverage I ran across two problems which are fixed in this PR.  The patches are small and should speak for themselves.

PR #191, opened 2019-11-20
BUG: use openat instead of open in test 15

keszybz:
On arm64, __NR_open is not defined, openat is always used. Let's use openat
instead, which is defined for architectures currently.

### DIFF
--- a/tests/15-basic-resolver.c
+++ b/tests/15-basic-resolver.c
@@ -55,7 +55,7 @@ int main(int argc, char *argv[])
 	unsigned int arch;
 	char *name = NULL;
 
-	if (seccomp_syscall_resolve_name("open") != __NR_open)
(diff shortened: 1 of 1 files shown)


pcmoore:
Hi @keszybz, thanks for bringing this up, we were supposed to have fixed this in a previous PR but it looks like something went wrong with the merge and the fix was dropped.  I've merged the original fix into the master branch via 1ecdddb2a5b61cf527d1f238f88a9d129239f87a; if you still see problems let us know.

For reference, the original fix was part of PR #182 and can be found here: https://github.com/pcmoore/misc-libseccomp/commit/35803ceb43c453762a3ab5177c5f8d5dbb813478

keszybz:
Yep, that should work too.

PR #196, opened 2020-01-07
RFE: add a section on release tarball signatures to README.md

pcmoore:

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,24 @@ CHANGELOG files.
 Those who are interested in contributing to the the project are encouraged to
 read the CONTRIBUTING in the top level directory.
 
+## Verifying Release Tarballs
+
+Before use you should verify the downloaded release tarballs and checksums
+using the detached signatures supplied as part of the release; the detached
+signature files are the "*.asc" files.  If you have GnuPG installed you can
+verify detached signatures using the following command:
+
+	# gpg --verify file.asc file
+
+At present, only the following keys are authorized to sign official libseccomp
+releases:
+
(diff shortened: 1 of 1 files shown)


drakenclimber:
Looks good.

pcmoore:
Merged via e5c4b5ee8086b936d28b080e5477950f632b4d43 into master and ea4dafc0f45d859f4ff157f1120e77d0f01f5473 into release-2.4.

pcmoore:
@drakenclimber instead of making release-X.Y backports a separate process, I'm thinking we should just have a discussion about the backport when we merge the PR/patch into master.  If we agree that it should be backported, let's just do it right then.

Thoughts?

drakenclimber:
I really like that idea.  Spreads the workload out throughout the release cycle rather than piling it on all at the end.

We just need to remember to do it every time a patch comes in :)

pcmoore:
> I really like that idea. Spreads the workload out throughout the release cycle rather than piling it on all at the end.

Cool, let's do that for here on out.  If we are reviewing a patch and we think it should be backported to a release-X.Y branch, let's call it out in the PR comments.

> We just need to remember to do it every time a patch comes in :)

Well, no one ever said this job was easy ;)

drakenclimber:
One slight disappointment with the github interface - I can add an issue to v2.5 or v2.4.3, but not both.  I was hoping to use that as one way to ensure issues are backported as necessary.

pcmoore:
Yeah, I had that same idea at one point in time and ran into the same disappointment :(

Issue #199, opened 2020-01-08
BUG: Travis builds are failing due to "pip install cython"


pcmoore:
When merging PRs today I noticed that Travis started failing due to problems installing Cython via pip:
* https://travis-ci.org/seccomp/libseccomp/builds/634085496

We need to fix this *soon*.

pcmoore:
I tried a quick fix of returning the build to the default Python version (v3.6 as of now), but with that we are now seeing a Python test failure.

Travis patch:
```
Author: Paul Moore <paul@paul-moore.com>
Date:   Wed Jan 8 00:47:03 2020 -0500

    travis: use the default Python version
    
    Signed-off-by: Paul Moore <paul@paul-moore.com>

diff --git a/.travis.yml b/.travis.yml
index da1cde5..723a885 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,6 @@ compiler:
   - gcc
 
 language: python
-python:
-  - "nightly"
 
 addons:
   coverity_scan:
```

Test failure: https://travis-ci.org/pcmoore/misc-libseccomp/builds/634088624
```
 batch name: 24-live-arg_allow
 test mode:  python
 test type:  live
Test 24-live-arg_allow%%001-00001 result:   FAILURE 24-live-arg_allow 1 ALLOW rc=159
```

drakenclimber:
Yuck - I agree this is high priority.  I'll try and help debug this today.

drakenclimber:
I have hacked around with this a little bit, and I don't like what I have found.  As best as I can tell, test 24 seems dependent upon what has been run before it.

I created a [branch](https://github.com/drakenclimber/libseccomp/commits/issues/fixtravisci2) and I have two patches in it.

Patch 1 - The above patch to disable the nightly python - Automated tests [fail](https://travis-ci.org/drakenclimber/libseccomp/builds/634413662?utm_source=github_status&utm_medium=notification) (due to test 24)
Patch 2 - Only run test 24 - Automated tests (i.e. test 24) [pass](https://travis-ci.org/drakenclimber/libseccomp/builds/634420945?utm_source=github_status&utm_medium=notification)

pcmoore:
> I have hacked around with this a little bit, and I don't like what I have found. As best as I can tell, test 24 seems dependent upon what has been run before it.

[...]

PR #201, opened 2020-01-09
BUG: use python 3.8 to prevent cython install failure

whereswaldon:
I was able to make the existing tests work by just switching back to the latest stable release of Python. I don't pretend to understand the problems that you were trying to debug @drakenclimber, but perhaps this will work reliably? I had a successful build on it [here](https://travis-ci.com/whereswaldon/libseccomp/builds/143730511).

Signed-off-by: Chris Waldon <chris.waldon@ibm.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ compiler:
 
 language: python
 python:
-  - "nightly"
+  - "3.8"
 
 addons:
   coverity_scan:


drakenclimber:
Awesome!!!  Thank you so much!

I didn't realize how tightly coupled Cython and our libseccomp code is with newer versions of Python.  I figured v3.6 would be new enough and blindly continued down that path.

whereswaldon:
It seems like it would be valuable to understand *why* we need to use really recent versions of Python, but that doesn't seem like a reason not to deploy this fix now.

pcmoore:
Hi @whereswaldon, thanks for the help and the patch!

Let's continue the broader discussion on how to resolve this over in #199.

pcmoore:
Let's close this out in favor of #202.

PR #204, opened 2020-01-24
syscalls: use gperf for lookups

giuseppe:
use gperf to generate a perfect hash to lookup syscall names.   It
improves significantly the complexity for seccomp_syscall_resolve_name.*
since it replaces the expensive strcmp for each syscall in the
database, with a lookup table.

The complexity for x86_64_syscall_resolve_num is not changed and it
uses the linear search, that is anyway less expensive than
seccomp_syscall_resolve_name.* as it uses an index for comparison
instead of doing a string comparison.

On my machine, calling 1000 seccomp_syscall_resolve_name_arch and
seccomp_syscall_resolve_num_arch over the entire syscalls DB passed
from ~0.45 sec to ~0.06s.

I've implemented it only for x86_64 as I've no access to other archs
for benchmarking the results.

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>


src/Makefile.am
```diff
@@ -67,5 +67,11 @@ libseccomp_la_CFLAGS = ${AM_CFLAGS} ${CODE_COVERAGE_CFLAGS} ${CFLAGS} \
 libseccomp_la_LDFLAGS = ${AM_LDFLAGS} ${CODE_COVERAGE_LDFLAGS} ${LDFLAGS} \
 	-version-number ${VERSION_MAJOR}:${VERSION_MINOR}:${VERSION_MICRO}
 
+EXTRA_DIST += arch-x86_64-syscalls.c
+CLEANFILES = arch-x86_64-syscalls.c
+
+arch-x86_64-syscalls.c: arch-x86_64-syscalls.c.perf
+	$(GPERF) -m 100 --null-strings --pic -ptCEG -S1 $< > $@
```
eparis: what is -p?  I don't see it in my help for gperf

giuseppe: good eye!  I assumed it was used by older versions (https://git.savannah.gnu.org/cgit/gperf.git/tree/src/options.cc#n934):
```
        case 'p':               /* Generated lookup function a pointer instead of int. */
          break;                /* This is now the default. */
```

but it is a no-op in all the releases and as far as the GNU perf git history goes, so I'll drop it

giuseppe:
@pcmoore could you please take a look?

src/arch-x86_64-syscalls.c.perf
```diff
@@ -0,0 +1,565 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library; if not, see <http://www.gnu.org/licenses>.
+ */
+#include <seccomp.h>
+#include <string.h>
+#include "arch.h"
+#include "arch-x86.h"
+%}
+struct arch_syscall_def_internal { int name; int num; int index; };
```
drakenclimber: I'm not a gperf expert, but I'm assuming that gperf couldn't build a table of `struct arch_syscall_def` like before.  That would remove the need for the weird `static` trickery being done in `x86_64_syscall_iterate()`.  Is my (mis)understanding correct?

src/arch-x86_64-syscalls.c.perf
```diff
@@ -0,0 +1,565 @@
...
+ *
+ * Return the syscall mapping at position @spot or NULL on failure.  This
+ * function should only ever be used internally by libseccomp.
+ *
+ */
+const struct arch_syscall_def *x86_64_syscall_iterate(unsigned int spot)
+{
+	unsigned int iter;
+        /* It is thread-unsafe.  It should be used only for testing.  */
+	static struct arch_syscall_def ret;
```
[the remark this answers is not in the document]
giuseppe: I've used the static trick only because this method is used only for the tests:

```
 * Return the syscall mapping at position @spot or NULL on failure.  This
 * function should only ever be used internally by libseccomp.
```

Otherwise it would be a breaking change

src/arch-x86_64-syscalls.c.perf
```diff
@@ -0,0 +1,565 @@
...
+
+	ret.name = NULL;
+	ret.num = -1;
```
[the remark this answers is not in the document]
giuseppe: sure, I'll change it

src/arch-x86_64-syscalls.c.perf
giuseppe: yeah unfortunately gperf doesn't build the same struct table.  The main issue is that name now is a `int` and we need `stringpool + wordlist[N].name` to get  a pointer back

pcmoore:
> @pcmoore could you please take a look?

Hi @giuseppe, thanks for you work on this.  I think this goes well with some other work I wanted to get done with libseccomp, but as I'm looking for the associated issue I realize that I never created one; the only reference I had outside my head was a discussion with @drakenclimber one day at lunch :/

Let me create that issue today and then we can talk further.

pcmoore:
> Hi @giuseppe, thanks for you work on this. I think this goes well with some other work I wanted to get done with libseccomp, but as I'm looking for the associated issue I realize that I never created one; the only reference I had outside my head was a discussion with @drakenclimber one day at lunch :/
> 
> Let me create that issue today and then we can talk further.

Okay, take a look at #207 and let me know what you think.

pcmoore:
Also, if someone wants to work on this and #207, I think we could consider this for the v2.5 milestone/release.

(This would make @drakenclimber and my lives *much* easier from a maintenance perspective, so anyone willing to work on this get's a large number of gold stars!)

giuseppe:
> Okay, take a look at #207 and let me know what you think.

thanks for writing this up.  I think it makes a lot of sense to not spread the same information among different files.

Would you be ok with having the .perf file as the single source instead of a .csv file?

It will be really easy to generalize the x86_64 code to handle multiple archs.  We will need to add a column for each arch, in the same way as you've proposed for the .csv file, so it will look more like:

```
+struct arch_syscall_def_internal { int name; int x86_64; int x86; ...; int index; };
```

We still need the index column as gperf doesn't keep the same ordering we provide.

If you prefer the csv file then we would need to generate the .perf file based on that.

The main disadvantage with using directly gperf is that the lookup by syscall number is still `O(n)` as we need to scan the entire table, although we are comparing a single int so it is significantly faster than `strcmp`.
Instead, if we'll use a .csv file we could easily build an additional lookup array for each arch so that the `syscall number -> name` can also be `O(1)`

pcmoore:
> Would you be ok with having the .perf file as the single source instead of a .csv file?

Well, a CSV file easy to manipulate with any number of tools, the *.perf file format looks a bit more convoluted.  Besides ...

> The main disadvantage with using directly gperf is that the lookup by syscall number is still O(n) as we need to scan the entire table, although we are comparing a single int so it is significantly faster than strcmp.
Instead, if we'll use a .csv file we could easily build an additional lookup array for each arch so that the syscall number -> name can also be O(1).

Is using the *.perf file format as the source even desirable given what you've written above?

The CSV file format is tool agnostic and gives us plenty of options for other "stuff", including this gperf optimization.  As of right now I think the CSV source format is our best option of the two.

giuseppe:
> > 
> 
> Is using the *.perf file format as the source even desirable given what you've written above?
> 
> The CSV file format is tool agnostic and gives us plenty of options for other "stuff", including this gperf optimization. As of right now I think the CSV source format is our best option of the two.

I was not sure as I didn't want to add yet another level of source generation.

I went forward and added a .csv file.  The perf file is generated from there.  I've also adapted the code for arm, as an example of how it is possible to move existing archs to gperf.

pcmoore:
I'm going to preface this by saying I know almost nothing about gperf at the moment, but it would be really nice if we could limit the amount of C code in the "syscalls.perf.template" file to close-to-nothing.  It looks like we would need `XXX_syscall_resolve_name`, `XXX_syscall_resolve_num`, and `XXX_syscall_iterate` and that is it, right?

I would also like to see the "syscalls.csv to syscalls.perf" generation commands encapsulated into a separate script instead of open-coded into the makefile.

It would also be nice to drop the quotes around the syscall names in the CSV file.  Let's also convert the "__PNR_xxx" references into something generic like just "PNR"; the conversion script can change that into the appropriate "__PNR_xxx" value.  Further, can we get rid of the explicit offsets in the CSV, e.g. "(__SCMP_NR_BASE + 140)"?  It seems like the conversion script could always assume an offset and generate the "(__SCMP_NR_BASE + xxx)" C code and it would be up to the ABI definition to define the offset; for ABIs like x86 that offset could be zero.

pcmoore:
Side note: you also need to find a way to get your code coverage numbers back up.  It might just be a symptom of the PR being a work-in-progress, but we've worked *way* too hard to have it the coverage numbers drop down into the 60% range.

giuseppe:
> I'm going to preface this by saying I know almost nothing about gperf at the moment, but it would be really nice if we could limit the amount of C code in the "syscalls.perf.template" file to close-to-nothing. It looks like we would need `XXX_syscall_resolve_name`, `XXX_syscall_resolve_num`, and `XXX_syscall_iterate` and that is it, right?
> 
> I would also like to see the "syscalls.csv to syscalls.perf" generation commands encapsulated into a separate script instead of open-coded into the makefile.
> 
> It would also be nice to drop the quotes around the syscall names in the CSV file. Let's also convert the "__PNR_xxx" references into something generic like just "PNR"; the conversion script can change that into the appropriate "__PNR_xxx" value. Further, can we get rid of the explicit offsets in the CSV, e.g. "(__SCMP_NR_BASE + 140)"? It seems like the conversion script could always assume an offset and generate the "(__SCMP_NR_BASE + xxx)" C code and it would be up to the ABI definition to define the offset; for ABIs like x86 that offset could be zero.

can we do these cleanups for the .csv file incrementally?  I've taken care of the other comments, and also the coverage seems to be again ok now.

pcmoore:
> can we do these cleanups for the .csv file incrementally?

I would like to see them all in the same PR.  The same goes for supporting all the arch/ABIs.

I'm not a big fan of incomplete work where there is just a promise to do it later ;)

giuseppe:
> I'm not a big fan of incomplete work where there is just a promise to do it later ;)

I wasn't even promising that :-)  I thought it was fine to just add it for x86_64 and other arches could be adapted later by whoever cares about them.

Can we just use the existing values to populate the csv for now?  I am not exactly sure how you'd like to rewrite them an it seems like an extra step that could be done separately.

Some arches, like s390, have some extra logic in `$ARCH_syscall_resolve_num`, and there are a series of if (num == VALUE) before iterating the table.  Is it just an optimization for the most used syscalls or is it something else?

giuseppe:
I've pushed a new version where I've adapted all the architectures

giuseppe:
@pcmoore does the last revision look good to you?

pcmoore:
Hi @giuseppe thanks for making the requested changes.  I'm scrambling a bit trying to wrap things up before I step away for a few days; I'll try to take a closer look when I get back.

giuseppe:
@pcmoore any chance to take a look at this PR?

pcmoore:
Not yet, hopefully I will be able to find some time this weekend.

src/arch-syscall-check.c
```diff
@@ -90,6 +94,15 @@ int main(int argc, char *argv[])
...
 
+		syscall_num = syscall_resolve_name(sys_name, OFFSET_ARCH(x86_64));
+		if (syscall_num >= 0) {
+			sys_name_tmp = syscall_resolve_num(syscall_num, OFFSET_ARCH(x86_64));
+			if (strcmp (sys_name, sys_name_tmp)) {
+				printf("syscall_resolve returns wrong data\n");
+				return 1;
+			}
+		}
+
```
pcmoore: You lost me here, what is the purpose of this addition?  Is it just to make sure `syscall_resolve_name()` and `syscall_resolve_num()` are working?

src/generate_syscalls_perf.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -xeuo pipefail
```
pcmoore: Since this is part of the build process, let's avoid any fancy bash settings and do any artifact cleanup as part of `make clean`.

src/generate_syscalls_perf.sh
```diff
@@ -0,0 +1,15 @@
...
+
+function cleanup {
+    rm -f syscalls.csv.tmp
+}
+trap cleanup EXIT
```
pcmoore: See above comment.

src/generate_syscalls_perf.sh
```diff
@@ -0,0 +1,15 @@
...
+
+grep -v ^# < syscalls.csv > syscalls.csv.tmp
```
pcmoore: Please put quotes around the `^#`, for example:
```
grep -v "^#" < syscalls.csv > syscalls.csv.tmp
```

src/generate_syscalls_perf.sh
pcmoore: FWIW, the reasoning for this is that sometimes it is nice to see the temporary build artifacts when things go wrong.

src/syscalls.h
```diff
@@ -0,0 +1,65 @@
...
+
+#define __SCMP_NR_OABI_SYSCALL_BASE    0x900000
+#define __SCMP_ARM_NR_BASE             0x0f0000
+/* NOTE: we currently only support the ARM EABI, more info at the URL below:
+ *       -> http://wiki.embeddedarm.com/wiki/EABI_vs_OABI */
+#if 1
+#define __SCMP_NR_BASE			0
+#else
+#define __SCMP_NR_BASE			__SCMP_NR_OABI_SYSCALL_BASE
+#endif
```
pcmoore: If we are going to move all of the per-ABI offsets into one file, we need to make sure the offset names are tied to their associated ABI.  As an example, the name `__SCMP_NR_BASE` is not a good choice here as the associated ABI is ambigious, but a name like `__SCMP_ARM_NR_BASE` is better as it clearly ties the offset define to the ARM ABI.

Does that make sense?

src/syscalls.perf.template
```diff
@@ -0,0 +1,82 @@
+%{
+/**
+ * Enhanced Seccomp x86_64 Syscall Table
```
pcmoore: This comment doesn't make much sense now :)

src/syscalls.perf.template
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * Copyright (c) 2012, 2020 Red Hat <pmoore@redhat.com>
+ * Author: Paul Moore <paul@paul-moore.com>
+ * gperf support: Giuseppe Scrivano <gscrivan@redhat.com>
```
pcmoore: This copyright block isn't quite right, it makes it seem as though I updated it in 2020, which I did not.  I would expect something like the following (the authors part is flexible):

```
 * Copyright (c) 2012 Red Hat <pmoore@redhat.com>
 * Copyright (c) 2020 Red Hat <gscrivan@redhat.com>
 * Authors:  Paul Moore <paul@paul-moore.com>, Giuseppe Scrivano <gscrivan@redhat.com>
```

src/arch-syscall-check.c
giuseppe: the lookup code is much bigger now, without such check that verifies every single syscall the code coverage will drop significantly.  The main reason for having it is code coverage, and since the lookup now is way more complex (at least for the generated code), it makes sense to test it for each syscall.

src/syscalls.perf.template
```diff
@@ -0,0 +1,82 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library; if not, see <http://www.gnu.org/licenses>.
+ */
+#include <seccomp.h>
+#include <string.h>
+#include "arch-x86.h"
+#include "arch-arm.h"
```
pcmoore: Why do we need to include only the arch/ABI headers for x86 and 32-bit ARM?

giuseppe: thanks for noticing it, it is a left-over from an old version.  We need only `#include "arch.h"`

pcmoore:
Hi @giuseppe, I want to apologize for the delay in reviewing the latest draft.  I added some comments inline, but I had a general comment regarding the CSV format that I thought was best shared here.

The current code uses the following format, note the cpp macro offsets and pseudo syscall number in the file:
```
access, 5, 21, (__SCMP_NR_BASE + 33), __PNR_access, (__SCMP_NR_BASE + 20), (__SCMP_NR_BASE + 20), (__SCMP_NR_BASE + 33), 33, 33, 33, 33, 33, (X32_SYSCALL_BIT + 21), 33
```

It would be nice if we could remove those and do something like the example below (I tried explaining this in a [comment](https://github.com/seccomp/libseccomp/pull/204#issuecomment-580506843) above, but I may have done a poor job):
```
access, 5, 21, 33, PNR, 20, 20, 33, 33, 33, 33, 33, 33, 21, 33
```
... in this example the build process, and or the gperf template code, automatically adds the necessary per-ABI offset and does the translation from the generic "PNR" to the appropriate pseudo syscall value.  Does that make sense?

giuseppe:
@pcmoore I took care of your comments and simplified the `.csv` file even further.  I've left only some values not precalculated (e.g. `0x0f0000 + 1`) as to me it seems easier to read.  If you want we can change these as well

src/arch-syscall-check.c
pcmoore: If this is purely for a testing, it should be part of a test and not in `arch-syscall-check.c`.  You could create a new test, maybe similar to 15-basic-resolver, that would exercise the syscall lookup APIs.  As you say, the code is much more complex now, a new test to exercise it would be a good idea.

configure.ac
```diff
@@ -119,6 +119,11 @@ AC_DEFINE_UNQUOTED([ENABLE_PYTHON],
 	[$(test "$enable_python" = yes && echo 1 || echo 0)],
 	[Python bindings build flag.])
 
+AC_CHECK_TOOL(GPERF, gperf)
+if test -z "$GPERF"; then
+	AC_MSG_ERROR(please install gperf)
```
pcmoore: This is a nit, but since we still need some other changes, how about the following (use square brackets) for the sake of consistency with the other `AC_MSG_ERROR()` call in configure.ac.
```suggestion
	AC_MSG_ERROR([please install gperf])
```

src/syscalls.csv
```diff
@@ -0,0 +1,468 @@
+# name, index, x86_64, arm, aarch64, mips64n32, mips64, mips, parisc, ppc64, ppc, s390, s390x, x32, x86. riscv64
```
pcmoore:
```suggestion
# name, index, x86_64, arm, aarch64, mips64n32, mips64, mips, parisc, ppc64, ppc, s390, s390x, x32, x86, riscv64
```

pcmoore: Also, thank you for adding RISC-V!  I wasn't going to ask you to do that, but that is nice to see :)

src/syscalls.csv
```diff
@@ -0,0 +1,468 @@
...
+accept, 3, 43, 285, 202, 42, 42, 168, 35, 330, 330, PNR, PNR, 0x40000000 + 43, PNR, 202
```
pcmoore: Why isn't the 0x40000000 offset included in the per-ABI offset for x32?

pcmoore:
Thanks @giuseppe for the update.

> I've left only some values not precalculated (e.g. 0x0f0000 + 1) as to me it seems easier to read. If you want we can change these as well.

It looks like the only values without the offset are the x32 syscalls, yes?  If so, I think those should be simplified as well.  For example, "0x40000000 + 43" should become just "43" in the CSV file.

src/syscalls.csv
pcmoore: Oh, one more thing - why do we need the second column, "index"?

pcmoore:
Ah, nevermind, I think I see what you were talking about @giuseppe:
```
usr26, 448, PNR, 0x0f0000 + 3, PNR, PNR, PNR, PNR, PNR, PNR, PNR, PNR, PNR, PNR, PNR, PNR
```
I guess I would be okay in this particular case, what do you think @drakenclimber?

pcmoore:
@drakenclimber we will need to update src/arch-syscall-validate when this gets merged, but it seems unfair to ask @giuseppe to do that ;)

In a perfect world src/arch-syscall-validate will generate a patch that we can apply directly to the syscall CSV file.  It shouldn't be that hard, and it will make updating the syscall table **soooo** much easier.

src/syscalls.csv
giuseppe: unfortunately gperf doesn't keep the same order when it generates its table.

I'll drop it from the csv and generate it

src/syscalls.csv
giuseppe: dropped and moved it as part of the arch definition

giuseppe:
@pcmoore thanks for the review!  I've addressed the comments and added a new test (hopefully the coverage won't be affected)

pcmoore:
Thanks @giuseppe this is looking good.  My last remaining question is why the coverage dropped a full 2%?

Looking at the latest [coveralls.io results](https://coveralls.io/builds/29049277) it appears that `arch-390.c` and `arch-s390x.c` had some lines "uncovered" in your run and it would be nice to know what changed and what we need to do to test those lines.

I realize code coverage is definitely not a guarantee of bug-free code, but we worked *really* hard to get that coverage number above 90% and I would *really* hate to see that number drop below 90% again :)

giuseppe:
it depends on the way the lookup function works now.

Before we were using something like:
```
	for (iter = 0; table[iter].name != NULL; iter++) {
		if (strcmp(name, table[iter].name) == 0)
			return table[iter].num;
	}
```

and it is easy to get full coverage here.

Now the lookup function is way more complex, and the only way to get full coverage is to lookup all the syscalls (and theorically even this could not be enough).

When I've added the test to `arch-syscall-check.c`, the coverage was fine.  My guess is that there is some sycall that is not being looked up.  Before it was easy to enumerate all the syscalls through `syscall_iterate` but now I've no access to it from the new test.

I'll check it better, might be some gaps in the syscalls numbers for s390/s390x that let the new test exit too early.

giuseppe:
> I'll check it better, might be some gaps in the syscalls numbers for s390/s390x that let the new test exit too early.

yes that was the reason.  I've changed the test and now the coverage is fine again.

pcmoore:
That's great, thanks for taking care of that @giuseppe.  My last nitpick is to say that the test `56-iterate-syscalls` should be named `56-basic-iterate_syscalls` to better fit with the rest of the naming convention but that is it from me - thanks for all your work on this.

@drakenclimber what do you think about this?

drakenclimber:
> @drakenclimber what do you think about this?

I have been loosely following along, and this seems like a really cool improvement.  Now that things have settled down, I'll give it a thorough look tomorrow.  Thanks!

pcmoore:
Thanks for renaming test 56 :)

src/generate_syscalls_perf.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+SYSCALLS_CSV=$1
+SYSCALLS_PERF_TEMPLATE=$2
+
+cat $SYSCALLS_CSV | grep -v '^#' | nl -ba -s, -v0 | \
+    sed -e '/^#/d' \
```
drakenclimber: Is this `sed -e '/^#/d` needed since you have the `grep -v  '^#'` above?

src/Makefile.am
```diff
@@ -31,20 +31,21 @@ SOURCES_ALL = \
...
+	arch-mips.h arch-mips.c \
+	arch-mips64.h arch-mips64.c \
+	arch-mips64n32.h arch-mips64n32.c \
+	arch-parisc.h arch-parisc.c arch-parisc64.c \
+	arch-ppc.h arch-ppc.c \
+	arch-ppc64.h arch-ppc64.c \
+	arch-riscv64.h arch-riscv64.c \
+	arch-s390.h arch-s390.c \
+	arch-s390x.h arch-s390x.c\
+	syscalls.c syscalls.perf.c
```
drakenclimber: (Sticking this comment here before I forget.) ;)

I noticed `arch-syscall-validate` raises several more differences for mips64 with this patch applied.  Probably more of any issue for @pcmoore and me but worth mentioning.

```
$ ./arch-syscall-validate -a mips64 ~/git/clean/upstream-torvalds/
--- mips64 [library]
+++ mips64 [system]
@@ -1,2 +1,340 @@
+accept	42
+accept4	293
+access	20
+acct	158
+add_key	239
+adjtimex	154
+afs_syscall	176
+alarm	37
+bind	48
...
```

src/syscalls.csv
```diff
@@ -0,0 +1,468 @@
+# name, x86_64, arm, aarch64, mips64n32, mips64, mips, parisc, ppc64, ppc, s390, s390x, x32, x86, riscv64
```
drakenclimber: I ran several architectures through `arch-syscall-validate`, but not all.  (Also note to self, riscv64 isn't in `arch-syscall-validate` yet.)  Most looked good, but see comment above about mips64.

tests/56-basic-iterate_syscalls.c
```diff
@@ -0,0 +1,84 @@
...
+	int n, iter = 0;
+
+	for (iter = init; iter < init + 1000; iter++) {
+		char *name;
+
+		name = seccomp_syscall_resolve_num_arch(arch, iter);
+		if (name == NULL)
+			continue;
+
+                n = seccomp_syscall_resolve_name_arch(arch, name);
```
drakenclimber: nitpick - spacing looks weird here.

tests/56-basic-iterate_syscalls.c
```diff
@@ -0,0 +1,84 @@
...
+	SCMP_ARCH_PARISC64,
+	SCMP_ARCH_RISCV64,
+	-1
+};
+
+static int test_arch(int arch, int init)
+{
+	int n, iter = 0;
+
+	for (iter = init; iter < init + 1000; iter++) {
```
drakenclimber: So just to make sure I understand this loop:

We're incrementing an index and seeing if that matches a valid syscall for this architecture.  If it is valid syscall, try to look up the syscall number (on line 64) given the name.

Is it okay that we're using the gperf table to resolve the number (in line 60) and the name (in line 64)?  I think it's okay, but I have a fear that we're using code under test to verify the code under test ;).

giuseppe: it is testing whether the resolved name given a syscall number returns the same syscall number.

The only reason why this test exists is to stress the lookup table :)

I've used the heuristic `init` to `init + 1000` just to make sure the search doesn't stop  if there is a gap in the numbering, e.g. if there are defined syscalls like: 1, 3, 4... make sure it keeps trying and not stop on the missing `2`.

In the first version, this check was part of `arch-syscall-validate` where I could look up directly into the syscalls table, but now that it is moved to `tests/`, I cannot access it directly.

tests/56-basic-iterate_syscalls.c
giuseppe: thanks, fixed

src/syscalls.csv
giuseppe: thanks for noticing it!

Indeed there were some issues in the generated csv file.

I've fixed and verified it for each architecture

src/generate_syscalls_perf.sh
giuseppe: correct, dropped it

src/Makefile.am
giuseppe: fixed in the last version

rhatdan:
@giuseppe @pcmoore @drakenclimber Is this ready to merge?  Would love to get this merged and a new release out, so we could take advantage.

drakenclimber:
Awesome!  Thanks for all the hard work @giuseppe 

giuseppe:
@pcmoore does the last version look good to you?

pcmoore:
Hi @giuseppe, can you explain the changes to src/arch-syscall-validate?

pcmoore:
> Hi @giuseppe, can you explain the changes to src/arch-syscall-validate?

Actually nevermind; I think I know what you were doing and that is just hiding some underlying problems with these changes and the syscalls table.  Let me fixup arch-syscall-validate properly so we can use it to generate the CSV from the kernel sources.  We are going to need to do that anyway, and it should avoid having to do what you did to hide problems in the CSV.

pcmoore:
@giuseppe @drakenclimber I think this is what we need for arch-syscall-validate:
* https://gist.github.com/pcmoore/c33e64237798f05b3ecafca19c5221e5

It's *really* crude but it works (hint: run it with the '-c' option and point it at a recent kernel source tree):
```
PS1> ./arch-syscall-validate -c ~/sources/linux/linux-linus
# libseccomp syscall table
#
# kernel: 5.6.0-rc6 (Fri, 20 Mar 2020 03:49:05 +0000)
#
#syscall,x86,x86_64,x32,arm,aarch64,mips,mips64,mips64n32,ppc,ppc64,riscv64,s390,s390x
accept,PNR,43,43,285,202,168,42,42,330,330,202,PNR,PNR
accept4,364,288,288,366,242,334,293,297,344,344,242,364,364
access,33,21,21,33,PNR,33,20,20,33,33,PNR,33,33
acct,51,163,163,51,89,51,158,158,51,51,89,51,51
add_key,286,248,248,309,217,280,239,243,269,269,217,278,278
adjtimex,124,159,159,124,171,124,154,154,124,124,171,124,124
```
@giuseppe if that looks okay to you, you can include that in your patchset as a second patch or I can add it when I merge your patch; either approach is fine with me.

@drakenclimber as you mentioned previously, we are going to really need to verify that we haven't screwed up the syscall tables with this change.  Thankfully this should be a one-time event, moving forward we can just generate a new CSV file for each release (and give the diff a quick sanity check of course).

giuseppe:
> > Hi @giuseppe, can you explain the changes to src/arch-syscall-validate?
> 
> Actually nevermind; I think I know what you were doing and that is just hiding some underlying problems with these changes and the syscalls table.

I think the issue was in the previous implementation for some archs.

For example s390 had:

```
int s390_syscall_resolve_name(const char *name)
{
       unsigned int iter;
       const struct arch_syscall_def *table = s390_syscall_table;

       /* XXX - plenty of room for future improvement here */

       if (strcmp(name, "accept") == 0)
               return __PNR_accept;
       if (strcmp(name, "accept4") == 0)
               return __PNR_accept4;
       else if (strcmp(name, "bind") == 0)
               return __PNR_bind;
       else if (strcmp(name, "connect") == 0)
               return __PNR_connect;
       ....
```

so the value returned by `*_resolve_name` was different than the value stored into the syscalls `table`.  With the new implementation there is not such per-arch customization and the value returned by `*_resolve_name` is the same as the value stored in the table.

`src/arch-syscall-validate` reads directly into the syscalls table and doesn't check for the real value returned by `_resolve_name`.

giuseppe:
> ```
> 
> ```
> 
> 
> @giuseppe if that looks okay to you, you can include that in your patchset as a second patch or I can add it when I merge your patch; either approach is fine with me.

@pcmoore I am fine with including your patch in the PR, but I think we still need the filters I've added initially to `arch-syscall-validate` for the comment I've left previously.

Are you fine with something like:
```
diff --git a/src/arch-syscall-validate b/src/arch-syscall-validate
index 1a508bf..4c21653 100755
--- a/src/arch-syscall-validate
+++ b/src/arch-syscall-validate
@@ -664,15 +664,21 @@ function gen_csv() {
        done
        printf "\n"
        # output the syscall csv details
+       for abi in $abi_list; do
+               dump_sys_$abi $1 > sys_$abi.list
+       done
        for sc in $sc_list; do
                printf "%s" $sc
                for abi in $abi_list; do
-                       num=$(dump_sys_$abi "$1" | grep "^$sc," | awk -F "," '{ print $2 }' )
+                       num=$(cat sys_$abi.list | grep "^$sc," | awk -F "," '{ print $2 }' )
                        [[ -z $num ]] && num="PNR"
                        printf ",%s" $num
                done
                printf "\n"
        done
+       for abi in $abi_list; do
+               rm sys_$abi.list
+       done
 }
```

on top of your patch?

With the patch above `arch-syscall-validate -c` takes ~14 seconds on my system, which I think is acceptable.

pcmoore:
> I think the issue was in the previous implementation for some archs.

Sorta, there is an issue in s390 and s390x where we aren't handling the syscall lookup correctly, see the discussion in #215 for more information.

This of course is a reminder that we really need to be able to specify ABI specific lookup functions, e.g. `x86_syscall_resolve_name()` and `x86_syscall_resolve_num()`.  The default raw hash table lookup function should work for most of the ABIs, but we will need special handling for some like x86, s390, and s390x.

@giuseppe do you mind if I spend some time hacking on your patch/PR?  I'm not sure what my weekend is looking like (COVID-19 has introduced a lot of uncertainty in the world), but I may have some time to work on this if that is okay with you.  Of course if you would prefer to fix it yourself, please do!

pcmoore:
> @pcmoore I am fine with including your patch in the PR, but I think we still need the filters I've added ...

See my previous comment, those filters are hiding a problem; we do not want them.

As I was reworking that patch (you had the right idea with only dumping the ABI lists once, but we really should use `mktemp` files like we do in other places) I realized that there is no reason we couldn't merge this now, independent of this PR.  I'm going to do a test build and if all looks good I'll merge it "soon".

@drakenclimber I'm going to be forgiveness on this one and merge the `arch-syscall-validate` without the proper two ACKs.  My reasoning is that we are the only ones that use this tool (and it is potentially blocking this PR) so if I mess it up too badly the pain will be limited :)

giuseppe:
> @giuseppe do you mind if I spend some time hacking on your patch/PR? I'm not sure what my weekend is looking like (COVID-19 has introduced a lot of uncertainty in the world), but I may have some time to work on this if that is okay with you. Of course if you would prefer to fix it yourself, please do!

of course I won't mind that :)

pcmoore:
Okay, please check out PR #223, that has my changes incorporated into with @giuseppe's changes.

giuseppe:
closing in favor of #223

Issue #193, opened 2019-11-26
BUG: inconsistent handling of ipc syscalls on s390 and s390x


cpaelzer:
Hi,
I was trying to push 2.4.2 to Ubuntu and after the few cleanups we had in 2.4.1->2.4.2 that looked pretty good at first. But When the extended checks kicked in I reliably found the systemd self-tests breaking on `i386` and `s390x`.
For what it is worth - they worked reliable on `x86-64`, `armhf`, `arm64` and `ppc64el`.

I know that there were related [systemd fixups](https://github.com/systemd/systemd/commit/4df8fe8415eaf4abd5b93c3447452547c6ea9e5f) but those are not strictly required since [our fix](https://github.com/seccomp/libseccomp/commit/e3647f5b6b52996bf30d0c2c1d1248e4182e1c1c) triggered by my chrony tests.

After further debugging of the testcase I found that of the many that systemd runs all kinds of activity around `shmat` seems to have changed.

The number resolves to 397 with old and new seccomp, but fails with version 2.4.2

    arch x86: SCMP_SYS(shmat) = 397
    ...
    Failed to add shmat() rule for architecture x86, skipping: Invalid argument

I was going back and forth and isolated the test more and more to now have reached a minimal test program that on `i386` as well as `s390x` reliable works/fails with `2.4.1`/`2.4.2`.

```
$ cat > test-seccomp-shmat.c << EOF
#include <linux/seccomp.h>

#include <errno.h>
#include <seccomp.h>
#include <stdio.h>

#include <sys/shm.h>

/*
 * Test issues with libseccomp 2.4.1 -> 2.4.2
 * Derived from systemd testcase test_memory_deny_write_execute_shmat
 * which fails to install shmat rules with 2.4.2 on i386 and s390x
 */

int main()
{
   int shmat_syscall = -1;
   int rc = -1;
   scmp_filter_ctx ctx;

   ctx = seccomp_init(SCMP_ACT_ALLOW);
   if (ctx == NULL)
       return -1;

   shmat_syscall = SCMP_SYS(shmat);
   printf("SCMP_SYS(shmat) = %d\n", shmat_syscall);

   rc = seccomp_rule_add_exact(ctx, SCMP_ACT_ERRNO(EPERM), shmat_syscall, 1, SCMP_A2(SCMP_CMP_MASKED_EQ, SHM_EXEC, SHM_EXEC));
   printf("Rule installed RC = %d\n", rc);

   return 0;
}
EOF
```

[...]

PR #51, opened 2016-11-02
RFE: refined header comment for `seccomp_arch_add`

guoger:
Error code `-EEXIST` is returned by this interface if specified arch
is already present. Added this information to header comment since it helps
developers to write code that safely ignores this ret code.

PR #205, opened 2020-01-27
Backports for v2.4.3

drakenclimber:
I looked through the commit log in master since v2.4.2 was released, and these commits were a good fit for the v2.4.3 release.

A few comments:
* I had a small collision backporting https://github.com/seccomp/libseccomp/commit/f1513f7eb2e9879007c23fbc6f0f71049bda7b07 because Test #51 (live user notifications) didn't exist in v2.4.  To keep things consistent between master and v2.4, I didn't renumber the basic load test, so there's no test #51 in the v2.4 branch
* I still need to work on the s390 problems outlined in Issue #193, but that will be a separate review
* Commit b233c4048bf85e5055e6becf2dd23efbf5436a99 didn't apply to the v2.4 branch because we haven't backported the SCMP_FLTATR_CTL_SSB and SCMP_ACT_NOTIFY features

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 # Travis CI configuration for libseccomp
 
-# https://docs.travis-ci.com/user/trusty-ci-environment
+# https://docs.travis-ci.com/user/reference/bionic
 # https://wiki.ubuntu.com/Releases
 
-dist: xenial
+dist: bionic
 sudo: false
 
 notifications:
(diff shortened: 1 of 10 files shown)


pcmoore:
This looks good to me, but for future reference I don't think commits "tests: introduce a simple seccomp_load() test as part of the non-live…" and "tools: fix clang problems in scmp_bpf_sim" were strictly necessary for a maintenance release.  The former is just to bump our test coverage and isn't critical for a preexisting release stream and the latter fixes a trivial memory leak in a short-lived developer tool which shouldn't affect production systems.

pcmoore:
Ooops, forgot my ACK.
```
```

PR #206, opened 2020-01-28
BUG: fix s390/s390x handling of shm* syscalls

drakenclimber:
Issue #193 exposed a few cases where libseccomp isn't completely supporting `shm*()` syscalls on s390/s390x.  This commit adds IPC support for `shmat()`, `shmdt()`, `shmget()`, and `shmctl()`.  Socket support for the aforementioned syscalls was unchanged and should still work as before.

I ran this through `arch-syscall-validate`, but I really don't have a way to test these changes.  @cpaelzer or others - do you have a way to test this?

### DIFF
--- a/src/arch-s390-syscalls.c
+++ b/src/arch-s390-syscalls.c
@@ -531,6 +531,14 @@ int s390_syscall_resolve_name(const char *name)
 		return __PNR_sendto;
 	else if (strcmp(name, "setsockopt") == 0)
 		return __PNR_setsockopt;
+	else if (strcmp(name, "shmat") == 0)
+		return __PNR_shmat;
+	else if (strcmp(name, "shmdt") == 0)
+		return __PNR_shmdt;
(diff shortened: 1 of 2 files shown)


drakenclimber:
@pcmoore, I think this is the last open issue for the v2.4.3 release.

pcmoore:
These changes look reasonable, but lets see if @cpaelzer can give them a test (I don't have access to s390[x] hardware).

As an aside, look at that code coverage :}

pcmoore:
At this point, let's just assume it all works on s390[x], if there is a bug we'll fix it.

Merged via 0a4c0300ebcc982cf419a4f7382ffc28e4792d2a, thanks @drakenclimber.

pcmoore:
Merged into the release-2.4 branch via 266e1b4c28cc5fd2e45de32c1f3509b94761f3e6.

PR #208, opened 2020-02-07
Q: possibly missing a AUDIT_ARCH_MIPSEL64 define

fangyaling:
The AUDIT_ARCH_MIPSEL64 define should be added for mips64 platform


pcmoore:
Hi @fangyaling, it looks like AUDIT_ARCH_MIPSEL64 was defined by the kernel headers going all the way back to v2.6, which is well before seccomp-bpf was introduced.  I don't see a reason why we need to explicitly add this to libseccomp, the proper thing to do is to make sure you have the kernel headers properly installed on your build system.

Are you running into a problem?  If so, can you provide some more detail about the problem?

fangyaling:
After confirm with my partner about the history of this define, that's right the define is no need here. Thanks for review!

Issue #210, opened 2020-03-04
BUG: test 53/55 failures/errors on aarch64


pcmoore:
It looks like we have a number test failures and errors on aarch64 with the current master branch; HEAD is 38f04da ("tests: add tests for the binary tree").

Test FAILUREs:
```
 test mode:  c
 test type:  bpf-valgrind
Test 53-sim-binary_tree%%331-00001 result:   FAILURE 53-sim-binary_tree rc=14
```

```
 test mode:  c
 test type:  basic
Test 55-basic-pfc_binary_tree%%001-00001 result:   FAILURE 55-basic-pfc_binary_tree.sh rc=1
```

Test ERRORs:
```
 test mode:  c
 test type:  bpf-sim
Test 53-sim-binary_tree%%001-00001 result:   ERROR 53-sim-binary_tree rc=14
...
<every 53-sim-binary_tree test in "c" mode ERRORs in the same manner>
```

```
 test mode:  python
 test type:  bpf-sim
Test 53-sim-binary_tree%%001-00001 result:   ERROR 53-sim-binary_tree rc=1
...
<every 53-sim-binary_tree test in "python" mode ERRORs in the same manner>
```

pcmoore:
@drakenclimber I'm not sure if you have access to an aarch64 system (maybe a Raspberry Pi running a 64-bit build?), but I'm going to assign this one to you for obvious reasons.  If you don't have access to an aarch64 system let me know and I'll look into this.

pcmoore:
I did a little digging just now and it appears that test `53-sim-binary_tree` isn't generating any input and is returning with error code 16/EBUSY when run standalone.

pcmoore:
I added some quick instrumentation (below):
```c
diff --git a/tests/53-sim-binary_tree.c b/tests/53-sim-binary_tree.c
index 2c7890e..c154f67 100644
--- a/tests/53-sim-binary_tree.c
+++ b/tests/53-sim-binary_tree.c
@@ -81,6 +81,7 @@ int main(int argc, char *argv[])
                                              SCMP_A0(SCMP_CMP_EQ, i));
                else
                        rc = seccomp_rule_add(ctx, SCMP_ACT_ERRNO(i), i, 0);
+               printf("PMD: i=%d, rc=%d\n", i, rc);
                if (rc < 0)
                        goto out;
        }
```
... which resulted in the following output:
```
# ./53-sim-binary_tree -p
PMD: i=0, rc=0
[...]

PR #211, opened 2020-03-05
BUG: fix test failures on aarch64 and other architectures

drakenclimber:
This commit is intended to address Issue #210 where tests 53 and 55 are failing on non-x86_64 architectures.  Both tests now utilize actual syscall names rather than fictitious syscall numbers.


src/arch-syscall-dump.c
```diff
@@ -127,6 +128,9 @@ int main(int argc, char *argv[])
 		case SCMP_ARCH_PPC64LE:
 			sys = ppc64_syscall_iterate(iter);
 			break;
+		case SCMP_ARCH_RISCV64:
+			sys = riscv64_syscall_iterate(iter);
+			break;
```
pcmoore: All the changes in `arch-syscall-dump.c` probably should be split out into a separate patch.  Yes, it is a bug fix, but it is a *different* bug that you are fixing.

tests/53-sim-binary_tree.c
```diff
@@ -49,38 +93,26 @@ int main(int argc, char *argv[])
...
+	for (i = 0; i < table_size; i++) {
 		/* arbitrarily make the filter more complex by filtering
 		 * on arguments for a few syscalls
 		 */
-		if (i == 10 || i == 53 || i == 61 || i == 255)
-			rc = seccomp_rule_add(ctx, SCMP_ACT_ERRNO(i), i, 1,
-					      SCMP_A0(SCMP_CMP_EQ, i));
+		if (table[i].syscall == SCMP_SYS(mprotect) ||
+		    table[i].syscall == SCMP_SYS(dup) ||
+		    table[i].syscall == SCMP_SYS(alarm))
```
pcmoore: Why not add an additional field to the syscall table to indicate that this syscall should have an argument added to the filter?

tests/53-sim-binary_tree.c
```diff
@@ -29,9 +29,53 @@
 
 #include "util.h"
 
-#define MAX_SYSCALL		(330)
+struct syscall_errno {
+	int syscall;
+	int error;
```
pcmoore: See my comments below about adding an additional field to signify that a given entry should add an argument to the filter.

tests/53-sim-binary_tree.py
```diff
@@ -36,11 +77,15 @@ def test(args):
...
     f.set_attr(Attr.CTL_OPTIMIZE, 2)
 
-    for i in range(330):
-        if (i == 10 or i == 53 or i == 61 or i == 255):
-            f.add_rule(ERRNO(i), i, Arg(0, EQ, i))
+    for entry in table:
+        print(entry)
+        if entry["syscall"] == "mprotect" or \
+           entry["syscall"] == "dup" or \
+           entry["syscall"] == "alarm":
```
pcmoore: My comments from the C variant of this test still apply here although the implementation will obviously be slightly different.

tests/55-basic-pfc_binary_tree.c
```diff
@@ -51,31 +76,31 @@ int main(int argc, char *argv[])
...
+	for (i = 0; i < table_size; i++) {
 		/* arbitrarily make the filter more complex by filtering
 		 * on arguments for a few syscalls
 		 */
-		if (i == 10 || i == 53 || i == 61 || i == 255)
-			rc = seccomp_rule_add(ctx, SCMP_ACT_ERRNO(i), i, 1,
-					      SCMP_A0(SCMP_CMP_EQ, i));
+		if (table[i].syscall == SCMP_SYS(write) ||
+		    table[i].syscall == SCMP_SYS(open) ||
+		    table[i].syscall == SCMP_SYS(pread64))
```
pcmoore: The test 53 comments also apply here.

src/arch-syscall-dump.c
drakenclimber: Good plan.  It looks like we missed RISCV support in a few of the tools.  I'll make a separate pull request for them soon

drakenclimber: Addressed in pull request #212

pcmoore:
Merged via HEAD 3fb32e45dbde214d7bd05da51fd0a9d6abfb8265, thanks!

PR #212, opened 2020-03-10
RFE: Add RISCV64 support to scripts

drakenclimber:
This patchset adds riscv64 support to arch-syscall-dump and arch-syscall-validate.

When working on other libseccomp issues, I noticed both of these scripts were missing riscv64 support.  It's possible there are other scripts that need to be updated as well.


pcmoore:
Looks good to me, merged at HEAD 00afcffb77141f8416ca156065c50eae471b8ad1.

PR #217, opened 2020-03-18
RFE: dynamically link the Python bindings against the main libseccomp library

### LINKED ISSUE #61: Can you make the python part of the library available via pip?

maxking:
This allows the Python package to dynamically link against libseccomp installed
on the host when installed so it can be shared via PyPI easily.

This is a draft because I still need to update makefile to point to right library to link against when developing instead of installing. Builds are probably gonna fail right now, but I am working on it.

For reference: https://cython.readthedocs.io/en/latest/src/tutorial/clibraries.html#dynamic-linking

Fix #61

### DIFF
--- a/src/python/setup.py
+++ b/src/python/setup.py
@@ -40,8 +40,8 @@ setup(
 	cmdclass = {'build_ext': build_ext},
 	ext_modules = [
 		Extension("seccomp", ["seccomp.pyx"],
-			# unable to handle libtool libraries directly
-			extra_objects=["../.libs/libseccomp.a"],
+			# Dynamically link against libseccomp.so.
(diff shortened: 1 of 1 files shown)


pcmoore:
Hi @maxking, there needs to be some discussion about PRs #216 and #217 in the main issue, #61.  We basically need to decide if we want to continue to support the Python bindings as part of the main libseccomp source distribution or if we want to split them out like we do for the golang bindings.

pcmoore:
It is also worth noting the number of failures this PR had in Travis, although some of those may be due to the Python bindings picking up the wrong libseccomp library (one of the reasons this needs discussion in the main issue).

Issue #118, opened 2018-05-19
BUG/Q: Multiple "not equal" comparisons on one argument in one call to seccomp_rule_add() do not generate a proper filter


varqox:
I want to add filter that allows fstat only if fd is not 0, 1 or 2. I cannot get it to work. As long as I want to restrict it only on one fd (e.g. 0) all works fine, but with more it does not. Here is some demonstrative code: 
```c
#include <seccomp.h>
#include <stdio.h>
#include <errno.h>

#define _GNU_SOURCE
#include <sys/syscall.h>
#include <unistd.h>

int main() {
	scmp_filter_ctx ctx = seccomp_init(SCMP_ACT_KILL);
	if (ctx == NULL) {
		perror("seccomp_init()");
		return 1;
	}

	seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(write), 0);

	// I want to disallow fstat() on stdin, stdout and stderr (for demonstrative purposes)
	seccomp_rule_add(ctx, SCMP_ACT_ERRNO(EPERM), SCMP_SYS(fstat), 1, SCMP_A0(SCMP_CMP_EQ, 0));
	seccomp_rule_add(ctx, SCMP_ACT_ERRNO(EPERM), SCMP_SYS(fstat), 1, SCMP_A0(SCMP_CMP_EQ, 1));
	seccomp_rule_add(ctx, SCMP_ACT_ERRNO(EPERM), SCMP_SYS(fstat), 1, SCMP_A0(SCMP_CMP_EQ, 2));

	seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(fstat), 3,
		SCMP_A0(SCMP_CMP_NE, 0),
		SCMP_A0(SCMP_CMP_NE, 1),
		SCMP_A0(SCMP_CMP_NE, 2));

	int stdout_fileno = 1;
	seccomp_export_pfc(ctx, stdout_fileno);

	seccomp_load(ctx);

	// Confirms that the rule was not generated properly
	write(1, ">>> 1\n", 6);
	syscall(SYS_fstat, 1, NULL);
	write(1, ">>> 2\n", 6);
	syscall(SYS_fstat, 4, NULL);
	write(1, ">>> 3\n", 6); // Should reach here

	return 0;
}
```

varqox:
Thank you for help.
[...]

PR #225, opened 2020-03-24
doc: improve the manpage for seccomp_rule_add()

pcmoore:
Add some text to clarify that you can only have one comparison per
syscall argument in a given rule.

### DIFF
--- a/doc/man/man3/seccomp_load.3
+++ b/doc/man/man3/seccomp_load.3
@@ -23,6 +23,16 @@ Loads the seccomp filter provided by
 .I ctx
 into the kernel; if the function
 succeeds the new seccomp filter will be active when the function returns.
+.P
+As it is possible to have multiple stacked seccomp filters for a given task
+(defined as either a process or a thread), it is important to remember that
+each of the filters loaded for a given task are executed when a syscall is
+made and the "strictest" rule is the rule that is applied.  In the case of
+seccomp, "strictest" is defined as the action with the lowest value (e.g.
+.I SCMP_ACT_KILL
+is "stricter" than
+.I SCMP_ACT_ALLOW
+).
 .\" //////////////////////////////////////////////////////////////////////////
 .SH RETURN VALUE
 .\" //////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


pcmoore:
@varqox do you think this additional manpage text helps clarify the problem you described in #118?

varqox:
@pcmoore that would definitely help. But that is only a symptom of the underlying problem:

AFAIK currently, there is no specification other than the source code of libseccomp that tells what will happen if I install **more than one** rule for a **single** syscall. Is it a union, intersection of both or maybe the latter is the actual one that takes effect or something else (based on what I observe right now on v2.4.2 it is something else indeed).

This is very important because user of the libseccomp has no idea how or even if it is possible to achieve something like (example here has no practical value, but there other real-world that look similar to it):
```
default rule is SCMP_ACT_KILL and
calling setgid() with argument from {0,1,2,6,7} is allowed and
calling setgid() with argument from {3,4,5} gives EPERM and
calling setgid() with argument not from {0,1,2,3,4,5,6,7} gives EINVAL
```
Could you provide us (the users of libseccomp) with some information of how multiple rules for one syscall combine and what are the limitations?

pcmoore:
Force pushed revision 2.

varqox:
Now all is clear. Thanks, @pcmoore.

pcmoore:
Great, glad that cleared things up.  @drakenclimber does this look okay to you?

pcmoore:
Last call @drakenclimber, object now or submit a PR to fix it later :)

drakenclimber:
One last minute thought/question:

While it's true that libseccomp does a union of two `seccomp_rule_add()` function calls, could a (newer) user be confused how this differs from the kernel's union of multiple seccomp filters on a process?  Obviously in that case, the strictest rule wins.

pcmoore:
> While it's true that libseccomp does a union of two seccomp_rule_add() function calls, could a (newer) user be confused how this differs from the kernel's union of multiple seccomp filters on a process?

Seems like a reasonable source of confusion, let me see what I can come up with to help.

pcmoore:
I just force pushed an update, let me know how that looks @drakenclimber.

drakenclimber:
Thanks!!!  That clears up my concern.

```Reviewed-by: Tom Hromatka <tom.hromatka@oracle.com>```

pcmoore:
Merged via 23bc6267f5705b18c3143c8a509d0a0d33e5e62e, thanks for the review everyone!

PR #215, opened 2020-03-11
RFE: enable s390x and ppc64le in Travis

pcmoore:
Add the s390x and ppc64le architectures in the Travis CI builds.  Hold this PR until the test errors have been resolved in Travis.

PR #226, opened 2020-03-24
BUG: update the s390 and s390x munge functions

drakenclimber:
The s390 and s390x munge functions were missing a few syscalls and that was causing failures in the s390x TravisCI run.

This is another step to resolving pull request #215

### DIFF
--- a/src/arch-s390.c
+++ b/src/arch-s390.c
@@ -44,12 +44,26 @@ int s390_syscall_resolve_name_munge(const char *name)
 		return __PNR_getsockopt;
 	else if (strcmp(name, "listen") == 0)
 		return __PNR_listen;
+	else if (strcmp(name, "msgctl") == 0)
+		return __PNR_msgctl;
+	else if (strcmp(name, "msgget") == 0)
+		return __PNR_msgget;
+	else if (strcmp(name, "msgrcv") == 0)
+		return __PNR_msgrcv;
+	else if (strcmp(name, "msgsnd") == 0)
+		return __PNR_msgsnd;
 	else if (strcmp(name, "recv") == 0)
 		return __PNR_recv;
 	else if (strcmp(name, "recvfrom") == 0)
 		return __PNR_recvfrom;
 	else if (strcmp(name, "recvmsg") == 0)
 		return __PNR_recvmsg;
+	else if (strcmp(name, "semctl") == 0)
+		return __PNR_semctl;
+	else if (strcmp(name, "semget") == 0)
+		return __PNR_semget;
+	else if (strcmp(name, "semtimedop") == 0)
+		return __PNR_semtimedop;
 	else if (strcmp(name, "recvmmsg") == 0)
 		return __PNR_recvmmsg;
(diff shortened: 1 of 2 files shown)


pcmoore:
Do we also need to add the following?
* semop
* shmat
* shmdt
* shmget
* shmctl

drakenclimber:
Good question.  I'll run them through the s390x VM to see if they're needed.  If so, I'll likely add them to the test to ensure they are properly verified in the future.

drakenclimber:
> Do we also need to add the following?
> 
>     * semop
>     * shmat
>     * shmdt
>     * shmget
>     * shmctl

Based upon [Arnd's commit](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=0d6040d4681735dfc47565de288525de405a5c99) that was a catalyst for much of this work, I would think they should be munged as well.

But I can't create a real-world test scenario where the munge of the above shm*() syscalls is required.  (At least on my s390x machine, that is.  Note that I don't have console access and grubby isn't accepting my attempts to change to a newer kernel, so I'm limited to a really old kernel.  I've tried hacking `/usr/include/asm/unistd.h` by hand to influence the behavior, but it hasn't affected any outcomes.)

Strangely, I can get test 37 to fail if the msg*() syscalls aren't munged, but the munging (or not munging) of shm*() doesn't affect the test.

I'm torn.  I _think_ you are correct that they should also be munged, but I can't prove it.  Thoughts, @pcmoore?

giuseppe:
please ignore my review, I've clicked by mistake while I was still checking if it is something I broke (and my knowledge is quite limited to make any real review)

pcmoore:
Doing the munge to a pseudo syscall should be harmless as the libseccomp code will Do The Right Thing and generate the necessary rules to handle both cases, where not doing the munge will cause only the new direct-wired syscall rule to be created.

That said, we should also take a closer look at s390/s390x to make sure that the `*_syscall_demux()` and `*_syscall_mux()` are also correct.

pcmoore:
> please ignore my review, I've clicked by mistake while I was still checking if it is something I broke (and my knowledge is quite limited to make any real review)

 :)

Thanks for looking, but this was broken long before your changes ;)

drakenclimber:
The various shm*() functions were already in both the [s390x](https://github.com/seccomp/libseccomp/blob/5abc3b1284faa7e3e63b6efed5300572ac0723ce/src/arch-s390x.c#L65) and [s390](https://github.com/seccomp/libseccomp/blob/master/src/arch-s390.c#L65) munge/demunge code, so they should be good to go.

`semop()` is not supported by [s390 or s390x](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/s390/kernel/syscalls/syscall.tbl).

I did add `semtimedop()` support to the s390 mux/demux functions as it had not yet been enabled.

pcmoore:
Looks good to me, merged at HEAD 0e3db6a4bbd15b151805e484fa52771b49eb9da7.  Thanks!

Issue #53, opened 2016-11-05
RFE: provide detached signatures of source tarballs


g4jc:
Hello,
First of all thank you for taking the time to sign your hashes! Unfortunately, there is a bit of a problem downstream. Most packaging systems require a GPG signature and the original source file, otherwise they fail. Hence in-line GPG causes a packaging dilemma for maintainers. 
As seen in the following script from a PKGBUILD below:

` |  ==> Making package: libseccomp 2.3.1-1 (Sat Nov  5 09:49:11 EDT 2016)
 |  ==> Retrieving sources...
 |    -> Found libseccomp-2.3.1.tar.gz
 |    -> Found libseccomp-2.3.1.tar.gz.SHA256SUM.asc
 |  ==> Validating source files with sha256sums...
 |      libseccomp-2.3.1.tar.gz ... Passed
 |      libseccomp-2.3.1.tar.gz.SHA256SUM.asc ... Passed
 |  ==> Verifying source file signatures with gpg...
 |      libseccomp-2.3.1.tar.gz.SHA256SUM ... SOURCE FILE NOT FOUND
 |  ==> ERROR: One or more PGP signatures could not be verified!
`

The problem is there is no  libseccomp-2.3.1.tar.gz.SHA256SUM for the packager to download. 

I see two possible ways of fixing this issue, the second being the most proper.

Solution 1) Provide an unsigned libseccomp-2.3.1.tar.gz.SHA256SUM, although this still does not allow one to run validpgpkeys script to verify the source tarball, it does verify the sha256 is correct which could be manually compared with the PKGBUILD creators. (Not optimal)

Solution 2) Consider creating actual detached signatures for your source tarballs as well. Debian is pushing for this good standard and a tutorial for it is available here: https://wiki.debian.org/Creating%20signed%20GitHub%20releases

Thank you for your time.

pcmoore:
Thanks, this is something I'll consider for our future releases.

pcmoore:
I just added a process document for creating new libseccomp releases, RELEASE_PROCESS.md, in commit 4e618d26c58e08b560d0ca894f3381c0dd5d514f and the release process calls for the creation of detached signatures.

Issue #37, opened 2016-05-25
BUG: seccomp_rule_add man page is incorrect


pcmoore:
As reported by @stevegrubb the seccomp_rule_add man page incorrectly states that the individual rule actions are triggered does _not_ match:

```
   SCMP_ACT_KILL
          The  thread will be killed by the kernel when it calls a syscall
          that does not match any of the configured seccomp filter rules.
```

All of the actions in the seccomp_rule_add man page need to be checked and it would probably be a good idea to check the other related man pages as well.

pcmoore:
Fixed in 2821a4a4bba940abc5d2a21c42b5bd83e156bd0f.

PR #36, opened 2016-05-16
BUG: fix typo in seccomp_merge.3

lucab:
Minor typo fix: s/assocated/associated/

PR #227, opened 2020-03-24
tests: Remove ipc syscalls from test 53-sim-binary_tree

drakenclimber:
The syscalls shmat, shmctl, and shmget are ipc syscalls on s390
and s390x.  This causes test #53 to fail due to the syscall number
not matching up with the values in the *.tests file.

This commit removes the aforementioned ipc syscalls since they are
being tested in tests #36 and #37.

This is another step toward resolving pull request #215.


pcmoore:
Merged via 5abc3b1284faa7e3e63b6efed5300572ac0723ce, thanks Tom.

PR #232, opened 2020-04-15
BUG: fix api update to check if SCMP_ACT_NOTIFY is supported

KentaTada:
The function of `_seccomp_api_update` needs to check if the action of `SCMP_ACT_NOTIFY` is supported at level 5 like `SCMP_ACT_LOG` at level 3.
Especially, the result of `sys_chk_seccomp_action` will affect `_support_seccomp_user_notif`

Signed-off-by: Kenta Tada <Kenta.Tada@sony.com>

### DIFF
--- a/include/seccomp.h.in
+++ b/include/seccomp.h.in
@@ -413,6 +413,9 @@ const struct scmp_version *seccomp_version(void);
  *      support for the SCMP_ACT_KILL_PROCESS action
  *  4 : support for the SCMP_FLTATR_CTL_SSB filter attrbute
  *  5 : support for the SCMP_ACT_NOTIFY action
+ *      support for using seccomp_notify_receive()
+ *      support for using seccomp_notify_respond()
+ *      support for using seccomp_notify_id_valid()
  *
  */
 unsigned int seccomp_api_get(void);
--- a/src/api.c
+++ b/src/api.c
@@ -115,7 +115,8 @@ static unsigned int _seccomp_api_update(void)
 		level = 4;
 
 	if (level == 4 &&
-	    sys_chk_seccomp_flag(SECCOMP_FILTER_FLAG_NEW_LISTENER) == 1)
+	    sys_chk_seccomp_flag(SECCOMP_FILTER_FLAG_NEW_LISTENER) == 1 &&
+	    sys_chk_seccomp_action(SCMP_ACT_NOTIFY) == 1)
 		level = 5;
 
 	/* update the stored api level and return */


KentaTada:
PTAL @pcmoore

pcmoore:
Hi @KentaTada, I apologize for the delay, let me look at this now ...

pcmoore:
I'm a little confused as to what problem this patch solves, why is the `SECCOMP_FILTER_FLAG_NEW_LISTENER` flag check not enough?  Do you have a reproducer you can share?

KentaTada:
@pcmoore 
I'm sorry for my poor explanation.
My understanding is that `seccomp_api_get()` not only confirms the supported API level but also changes the value of internal flag like `_support_seccomp_action_log` when that action is supported.
For example, `_seccomp_api_update` will execute `sys_chk_seccomp_action(SCMP_ACT_LOG)` when `seccomp_api_get()` is executed.
https://github.com/seccomp/libseccomp/blob/ff2a538cfcfbd1d10c4b460d74d725ac866136f4/src/api.c#L110
`sys_chk_seccomp_action(SCMP_ACT_LOG)` changes the value of `_support_seccomp_action_log` in acodance with the result of below
https://github.com/seccomp/libseccomp/blob/ff2a538cfcfbd1d10c4b460d74d725ac866136f4/src/system.c#L155-L157

But the value of `_support_seccomp_user_notif` is not changed even if the API level is 5.
So, this commit makes the behavior same.

Specifically the value of `_support_seccomp_user_notif` affects the result of some APIs like `seccomp_notify_receive()`.
https://github.com/seccomp/libseccomp/blob/ff2a538cfcfbd1d10c4b460d74d725ac866136f4/src/system.c#L385

I found this issue when I wrote the sample programs to pass the file descriptor of seccomp notification via the UNIX domain socket using SCM_RIGHT between client and server.
Below is related behavior.
1.  Client initializes the seccomp context using seccomp_init()
2.  Client sends the file descriptor which is created by seccomp_notify_fd() to the server.
3.  Server receives the file descriptor from client.
4.  Server executes seccomp_apt_get() to confirm the server side 's API level.
5.  Server handles the client's notification using the notification API like `seccomp_notify_receive()`.
 
If  `seccomp_api_get()` does not change the value of `_support_seccomp_user_notif`, server cannot execute the notification API like `seccomp_notify_receive()`.

pcmoore:
Thanks for the detailed explanation @KentaTada!  Now I understand the problem, and why your fix makes sense.

Could you also update the comment block include/seccomp.h.in above seccomp_api_get() to indicate that API level 5 also supports the various `seccomp_notify_*(...)` functions (please explicitly list them all)?  That's another bug that we should fix for the upcoming v2.5.0 release.  If you aren't comfortable with that, let me know and I'll take care of it.

pcmoore:
@drakenclimber any objections to this fix or the seccomp.h.in comment block update?  Can you think of anything else we're missing here?

KentaTada:
> Thanks for the detailed explanation @KentaTada! Now I understand the problem, and why your fix makes sense.
> 
> Could you also update the comment block include/seccomp.h.in above seccomp_api_get() to indicate that API level 5 also supports the various `seccomp_notify_*(...)` functions (please explicitly list them all)? That's another bug that we should fix for the upcoming v2.5.0 release. If you aren't comfortable with that, let me know and I'll take care of it.

OK. I updated the comment of `seccomp_api_get()` about three APIs(seccomp_notify_receive(), seccomp_notify_respond(), seccomp_notify_id_valid()) which needs to enable `_support_seccomp_user_notif`.

drakenclimber:
Really good find, @KentaTada.  Thanks!

@pcmoore I think this change looks good, but I was unable to find similar handling for `SCMP_ACT_KILL_PROCESS` and its corresponding flag `_support_seccomp_kill_process`.  We don't need to fix it in this commit, but I think it's also a bug.  Looking through the comment header for `seccomp_api_get()`, it appears we need to update the level3 handling for kill process.

`Acked-by: Tom Hromatka <tom.hromatka@oracle.com.`

KentaTada:
@drakenclimber Thank you for confirming. 
I don't know the reason why `_support_seccomp_kill_process` is not updated.
But I think it is just a bug too and it should be modified as same as this issue.
Could you also review https://github.com/seccomp/libseccomp/pull/237 I created?

drakenclimber:
> @drakenclimber Thank you for confirming.
> I don't know the reason why `_support_seccomp_kill_process` is not updated.
> But I think it is just a bug too and it should be modified as same as this issue.
> Could you also review #237 I created?

Thanks so much for picking up that fix as well.  I _really_ appreciate the help!

pcmoore:
@drakenclimber are you okay with this revision?

drakenclimber:
> @drakenclimber are you okay with this revision?

Yep, looks good to me.

pcmoore:
Merged via ab57460fa781a1ef9e3860f1314a3cb8738f94ed, thanks everyone!

Issue #97, opened 2017-10-07
RFE: more precisely defined error return values 


topimiettinen:
Looking at libseccomp error return values about system calls, it is not possible to determine which of the following reasons caused the errors:

1. syscall doesn't exist on some arch 
2. syscall cannot be matched on some arch (because it's multiplexed, think socket/socketcall)
3. other error cases

When constructing the seccomp filter, the caller may consider some of these reasons fatal but not the others, so more detailed error information (wider set of error values than just `-EINVAL`) would be needed.

See also [systemd PR 6952](https://github.com/systemd/systemd/pull/6952).

pcmoore:
Hi @topimiettinen, sorry it has taken so long to get to this, but I think it's time we fix this.

@drakenclimber this one is going to be a doozy.  All of the libseccomp APIs should have manpages at this point (if they don't we need to create an issue for that), with all of the manpages having some hand-wavy "negative values on error" comments in the RETURN VALUE section.  I think we need to do the following:
* manually audit each API call to generate a list of possible return values
* decide if these return values make sense, modify the code if they don't
* document each possible return value in the associated manpage with a brief explanation of what the error code indicates

Thoughts?

drakenclimber:
> Hi @topimiettinen, sorry it has taken so long to get to this, but I think it's time we fix this.
> 
> @drakenclimber this one is going to be a doozy. All of the libseccomp APIs should have manpages at this point (if they don't we need to create an issue for that), with all of the manpages having some hand-wavy "negative values on error" comments in the RETURN VALUE section. I think we need to do the following:
> 
>     * manually audit each API call to generate a list of possible return values
> 
>     * decide if these return values make sense, modify the code if they don't
> 
>     * document each possible return value in the associated manpage with a brief explanation of what the error code indicates
> 
> 
> Thoughts?

Dang.... I grudgingly agree with everything you wrote above, especially the effort required :).  Getting the return codes right, and then documenting them will be a really big effort.

And yes, while not glamorous work, I think it's critical.  I have been working on some cgroup stuff and we recently ran across a container implementation that completely misunderstood a cgroup feature... but there was no definitive document in the kernel or anywhere else to properly guide the way, so they made it work as best they could.

topimiettinen:
It's great to see activity on this issue! I'm not against a thorough review, but the original request was limited only to be able to distinguish different failure modes, which is somewhat orthogonal. The review would certainly help, even prerequisite to a degree I think.

[...]

PR #241, opened 2020-05-30
RFE: cleanup the API return values and document them the in the manpages

pcmoore:
This has already been covered in detail in issue #97 so I won't repeat the details here.  In summary, this PR establishes the APIs error codes as part of the API from here moving forwards.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -610,7 +610,7 @@ API int seccomp_export_bpf(const scmp_filter_ctx ctx, int fd)
 	rc = write(fd, program->blks, BPF_PGM_SIZE(program));
 	gen_bpf_release(program);
 	if (rc < 0)
-		return -errno;
+		return -ECANCELED;
 
 	return 0;
 }
--- a/src/api.c
+++ b/src/api.c
@@ -449,7 +449,7 @@ API int seccomp_rule_add_array(scmp_filter_ctx ctx,
 	if (rc < 0)
 		return rc;
 	if (action == col->attr.act_default)
-		return -EPERM;
+		return -EACCES;
 
 	return db_col_rule_add(col, 0, action, syscall, arg_cnt, arg_array);
 }
@@ -498,7 +498,7 @@ API int seccomp_rule_add_exact_array(scmp_filter_ctx ctx,
 	if (rc < 0)
 		return rc;
 	if (action == col->attr.act_default)
-		return -EPERM;
+		return -EACCES;
 
 	if (col->filter_cnt > 1)
 		return -EOPNOTSUPP;
--- a/src/api.c
+++ b/src/api.c
@@ -653,9 +653,9 @@ API int seccomp_export_bpf(const scmp_filter_ctx ctx, int fd)
 	if (_ctx_valid(ctx))
 		return _rc_filter(-EINVAL);
 
-	program = gen_bpf_generate((struct db_filter_col *)ctx);
-	if (program == NULL)
-		return _rc_filter(-ENOMEM);
+	rc = gen_bpf_generate((struct db_filter_col *)ctx, &program);
+	if (rc < 0)
+		return _rc_filter(rc);
 	rc = write(fd, program->blks, BPF_PGM_SIZE(program));
 	gen_bpf_release(program);
 	if (rc < 0)
--- a/src/api.c
+++ b/src/api.c
@@ -94,6 +94,27 @@ static int _rc_filter(int err)
 }
 
 /**
+ * Filter the system error codes we send back to callers
+ * @param col the filter collection
+ * @param err the error code
+ *
+ * This is similar to _rc_filter(), but it first checks the filter attribute
+ * to determine if we should be filtering the return codes.
+ *
+ */
+static int _rc_filter_sys(struct db_filter_col *col, int err)
+{
+	/* pass through success values */
+	if (err >= 0)
+		return err;
+
+	/* pass the return code if the SCMP_FLTATR_API_SYSRAWRC is true */
+	if (db_col_attr_read(col, SCMP_FLTATR_API_SYSRAWRC))
+		return err;
+	return -ECANCELED;
+}
+
+/**
  * Validate a filter context
  * @param ctx the filter context
  *
@@ -355,12 +376,14 @@ API int seccomp_arch_remove(scmp_filter_ctx ctx, uint32_t arch_token)
 API int seccomp_load(const scmp_filter_ctx ctx)
 {
 	struct db_filter_col *col;
+	bool rawrc;
 
 	if (_ctx_valid(ctx))
 		return _rc_filter(-EINVAL);
 	col = (struct db_filter_col *)ctx;
 
-	return _rc_filter(sys_filter_load(col));
+	rawrc = db_col_attr_read(col, SCMP_FLTATR_API_SYSRAWRC);
+	return _rc_filter(sys_filter_load(col, rawrc));
 }
 
 /* NOTE - function header comment in include/seccomp.h */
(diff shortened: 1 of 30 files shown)


pcmoore:
Forced a push to resolve the merge conflict against master.

src/api.c
```diff
@@ -610,7 +610,7 @@ API int seccomp_export_bpf(const scmp_filter_ctx ctx, int fd)
 	rc = write(fd, program->blks, BPF_PGM_SIZE(program));
 	gen_bpf_release(program);
 	if (rc < 0)
-		return -errno;
+		return -ECANCELED;
```
drakenclimber: I'm wondering if we should consider adding a log/debug feature that could capture the `errno` data we're now throwing away.  It may prove help debugging when a customer files a bug against us.

src/gen_pfc.c
```diff
@@ -469,18 +469,17 @@ static int _gen_pfc_arch(const struct db_filter_col *col,
...
 {
-	int rc = 0;
 	int newfd;
 	unsigned int iter;
 	FILE *fds;
 
 	newfd = dup(fd);
 	if (newfd < 0)
-		return errno;
+		return -ECANCELED;
```
drakenclimber, line 478: Random low-priority comment - when I was cscoping errno, I noticed some of our functions in util.c and scmp_bpf_disasm.c sometimes return `-errno` and sometimes `errno`.  The doxygen comments of the functions imply they all should be returning a negative error number.

src/system.c
```diff
@@ -407,6 +409,6 @@ int sys_notify_id_valid(int fd, uint64_t id)
 		return -EOPNOTSUPP;
 
 	if (ioctl(fd, SECCOMP_IOCTL_NOTIF_ID_VALID, &id) < 0)
-		return -errno;
+		return -ENOENT;
```
drakenclimber, line 412: I didn't expect to see an `-ENOENT` error in this patch, but I think it's the right change to make for this ioctl.

src/gen_bpf.c
```diff
@@ -2254,6 +2251,9 @@ struct bpf_program *gen_bpf_generate(const struct db_filter_col *col)
 	struct bpf_state state;
 	struct bpf_program *prgm;
 
+	if (col->filter_cnt == 0)
+		return NULL;
```
drakenclimber, line 2255: Returning NULL here will lead to us returning `-ENOMEM`.  The comment in `_rc_filter()` says this about `-ENOMEM`:

```
/* NOTE: unable to allocate enough memory to perform the
*       requested operation */
```
I am afraid this will cause confusion for some users.  Should we update the above comment?  Or is it fair to return `-EINVAL` because the user tried to build a bpf program with no valid filters in it?

src/gen_pfc.c
```diff
@@ -289,7 +289,7 @@ static int _get_bintree_syscall_num(const struct pfc_sys_list *cur,
 	}
 
 	if (cur == NULL)
-		return -EINVAL;
+		return -EFAULT;
```
drakenclimber, line 292: Unrelated to this change - I noticed that `_gen_pfc_arch()`can return a negative return code, but we're ignoring the return code entirely in `gen_pfc_generate()`.  Not a big deal currently, but probably a nitpick worth fixing someday

drakenclimber: Oh, and `-EFAULT` isn't in `_rc_filter()`, but as mentioned in the comment above, these errors are being suppressed entirely.  So not a big deal at the moment.

src/api.c
```diff
@@ -210,7 +255,7 @@ API int seccomp_reset(scmp_filter_ctx ctx, uint32_t def_action)
 	if (ctx == NULL || db_col_action_valid(NULL, def_action) < 0)
```
drakenclimber, line 255: Argh - github won't let me comment on code outside of the area that was changed.

Do you need a `_rc_filter()` call on line 230 of src/api.c in `seccomp_api_set()`?  It's returning `-EINVAL` without being checked

src/api.c
```diff
@@ -210,7 +255,7 @@ API int seccomp_reset(scmp_filter_ctx ctx, uint32_t def_action)
...
 		return -EINVAL;
```
drakenclimber, line 256: Do you want to call `_rc_filter()` here as well?

pcmoore: Good catch, thanks.  I did a second check and caught a few others in src/api.c.  This should be fixed in the next force push.

src/api.c
pcmoore: Ha, `seccomp_api_set()` was the other function I caught went I double checked src/api.c :)

Thanks for noticing this, should be fixed in the next force push.

src/gen_bpf.c
pcmoore: This is tricky as currently `gen_bpf_generate(...)` doesn't really offer a lot of granularity when it comes to return value: it's either a valid pointer or NULL.  If there is any hope, it is that `_gen_bpf_build_bpf(...)` does return an error code and not simply a pointer; we could modify `gen_bpf_generate(...)` to return the `bpf_program` pointer via an argument and return a more useful return code ... assuming we got back and audit all the return codes in "src/gen_bpf.c".

src/system.c
pcmoore: Not sure what to say in response here other than "yup" :)

src/gen_pfc.c
pcmoore: Hmm, good point.  We're not quite extending the same return value promise to these tools (yet), but we should fix that.  I'll add a patch to the next force push.

pcmoore: Now that I'm looking at it, scmp_bpf_disasm.c looks correct, it's just the comments that are wrong.  For the API we want to return negative values on error, but for the terminal apps we want to return a positive errno value.

Does that make sense, or am I crazy?

src/gen_pfc.c
pcmoore: > I noticed that _gen_pfc_arch()can return a negative return code, but we're ignoring the return code entirely in gen_pfc_generate(). Not a big deal currently, but probably a nitpick worth fixing someday.

Seems related to the discussion above, I'm going to mark this comment resolved in favor of discussion it up there.

> Oh, and -EFAULT isn't in _rc_filter() ...

Well, since the default is `-EFAULT` it doesn't need to be :)

>

src/api.c
pcmoore: I see your point, I've got an idea, let me see what I can come up with ... stay tuned.

src/gen_bpf.c
pcmoore: I just added another patch to this PR which should address this; I'm going to mark this as resolved, if there are any issues let's discuss them against the newly added code.

src/api.c
```diff
@@ -653,9 +653,9 @@ API int seccomp_export_bpf(const scmp_filter_ctx ctx, int fd)
 	if (_ctx_valid(ctx))
 		return _rc_filter(-EINVAL);
 
-	program = gen_bpf_generate((struct db_filter_col *)ctx);
-	if (program == NULL)
-		return _rc_filter(-ENOMEM);
+	rc = gen_bpf_generate((struct db_filter_col *)ctx, &program);
```
drakenclimber, line 656: Thanks for making this change.  I think this will really help in debugging

src/gen_pfc.c
drakenclimber: Yes, I agree with your analysis.  The use of positive errors on a terminal has always been a pet peeve of mine because it seems so... error prone :(.

Let's fix up the comments some time, but it's obviously not critical or related to this work.

src/api.c
pcmoore: As one potential option, I just added a patch to this PR, "api: add the SCMP_FLTATR_API_SYSRAWRC filter attribute".  I'm going to mark this as resolved and we can continue the conversations in the comments or as review comments in the relevant code.

src/gen_pfc.c
pcmoore: > ... a pet peeve of mine because it seems so... error prone :(

:)

> Let's fix up the comments some time, but it's obviously not critical or related to this work.

I thought I fixed the comments, but I'm sure I missed a few ... can you point me in the right direction?

pcmoore:
As mentioned in the review, I added a new patch which introduces a new filter attribute, `SCMP_FLTATR_API_SYSRAWRC` which I think should allow callers to chose between a stable set of return codes, and the full set of return codes provided by the system.  Unfortunately I wish we didn't have to choose, but given the wide range of kernels, libcs, etc. I'm not sure how we do that.  If anyone has any ideas I would *love* to hear them.

I will admit to somewhat rushing the patches today as I wanted to get them out for review; I'll give them a better look tomorrow, but I wanted to see if the filter attribute idea seemed reasonable.

pcmoore:
> I briefly looked through the RAWRC patch. It seems like a solid solution to me although I would like to think about it a bit more.

I understand, it is far from perfect but I'm struggling to come up with something better.  I'm always open to ideas if anyone has any.  @drakenclimber I'm going to hold off on adding your ACK and pushing until I hear an "I can't think of anything better, I'm okay with this".

I also just did a force push which adds some tests and rebases against our current master branch.

pcmoore:
FYI, for some reason Travis canceled a could of the PR runs around this time so something must of been wonky with the infrastructure - I think we can ignore the current failed Travis run for now.

drakenclimber:
> I understand, it is far from perfect but I'm struggling to come up with something better. I'm always open to ideas if anyone has any. @drakenclimber I'm going to hold off on adding your ACK and pushing until I hear an "I can't think of anything better, I'm okay with this".

Same here.  I think this is a good first step until we think of something better.  I'm good with the RAWRC changes.

pcmoore:
Thanks @drakenclimber, I think we're on the same page here.  Although one thing to keep in mind is that API changes such as the filter attribute are forever - we don't get to back them out.  That said, it's a simple enough API that it should give us some wiggle room if needed.

I've rebased this PR on top of master, added your ACKs, and force pushed.  Assuming Travis runs clean I'll go ahead and merge the PR.  If you have any second thoughts we can always revert the changes before we release v2.5.0.

PR #242, opened 2020-05-30
RFE: force an API level update when necessary

pcmoore:
We can't always rely on callers calling seccomp_api_get() before
using any API level gated functionality so let's force an API level
update in a few key places.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -192,6 +192,9 @@ API int seccomp_api_set(unsigned int level)
 /* NOTE - function header comment in include/seccomp.h */
 API scmp_filter_ctx seccomp_init(uint32_t def_action)
 {
+	/* force a runtime api level detection */
+	_seccomp_api_update();
+
 	if (db_col_action_valid(NULL, def_action) < 0)
 		return NULL;
 
@@ -531,6 +534,9 @@ API int seccomp_rule_add_exact(scmp_filter_ctx ctx,
 API int seccomp_notify_alloc(struct seccomp_notif **req,
 			     struct seccomp_notif_resp **resp)
 {
+	/* force a runtime api level detection */
+	_seccomp_api_update();
+
 	return sys_notify_alloc(req, resp);
 }
 
@@ -559,6 +565,9 @@ API int seccomp_notify_respond(int fd, struct seccomp_notif_resp *resp)
 /* NOTE - function header comment in include/seccomp.h */
 API int seccomp_notify_id_valid(int fd, uint64_t id)
 {
+	/* force a runtime api level detection */
+	_seccomp_api_update();
+
 	return sys_notify_id_valid(fd, id);
 }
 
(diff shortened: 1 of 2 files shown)


pcmoore:
This PR revealed another problem: because TSYNC and NEW_LISTENER send information back via the `seccomp(...)` return value, they can not be used at the same time.  It seems like the right thing to do is to always set the NEW_LISTENER flag (as long as the API level supports it) unless the caller explicitly sets TSYNC.

Working on a patch now...

pcmoore:
I just pushed another patch on this PR which should resolve the problem.

pcmoore:
Looking at the kernel code, I noticed the following commit:

```
commit 51891498f2da78ee64dfad88fa53c9e85fb50abf
Author: Tycho Andersen <tycho@tycho.ws>
Date:   Wed Mar 4 11:05:17 2020 -0700

    seccomp: allow TSYNC and USER_NOTIF together
    
    The restriction introduced in 7a0df7fbc145 ("seccomp: Make NEW_LISTENER and
    TSYNC flags exclusive") is mostly artificial: there is enough information
    in a seccomp user notification to tell which thread triggered a
    notification. The reason it was introduced is because TSYNC makes the
    syscall return a thread-id on failure, and NEW_LISTENER returns an fd, and
    there's no way to distinguish between these two cases (well, I suppose the
    caller could check all fds it has, then do the syscall, and if the return
    value was an fd that already existed, then it must be a thread id, but
    bleh).
    
    Matthew would like to use these two flags together in the Chrome sandbox
    which wants to use TSYNC for video drivers and NEW_LISTENER to proxy
    syscalls.
    
    So, let's fix this ugliness by adding another flag, TSYNC_ESRCH, which
    tells the kernel to just return -ESRCH on a TSYNC error. This way,
    NEW_LISTENER (and any subsequent seccomp() commands that want to return
    positive values) don't conflict with each other.
    
    Suggested-by: Matthew Denton <mpdenton@google.com>
    Signed-off-by: Tycho Andersen <tycho@tycho.ws>
    Link: https://lore.kernel.org/r/20200304180517.23867-1-tycho@tycho.ws
    Signed-off-by: Kees Cook <keescook@chromium.org>
```

... which I expect will be part of Linux v5.7 when it is released *soon*.  Unfortunately, I think this is something we should add for the v2.5.0 release, what do you think @drakenclimber ?

drakenclimber:
> ... which I expect will be part of Linux v5.7 when it is released _soon_. Unfortunately, I think this is something we should add for the v2.5.0 release, what do you think @drakenclimber ?

Agreed.  I created Issue #243 to track it.

drakenclimber:
Looks good to me.

pcmoore:
> Agreed. I created Issue #243 to track it.

Okay, I'll see what I can do.  Fingers crossed that will be the last hurdle for v2.5.0.

> Looks good to me.

Thanks for the quick review, merged at HEAD 777b647a59a089612c4eb354c67ecc0bbe70cb9a.

PR #246, opened 2020-06-11
BUG: pick better names for some legacy terminology

pcmoore:
I've never wanted my code to be either a reminder or a source of
hurt for others, but it is possible that some older terminology used
in this repository may do just that.  That's a bug we need to fix,
and hopefully this patch does just that.


src/system.c
```diff
@@ -35,9 +35,9 @@
 #include "gen_bpf.h"
 #include "helper.h"
 
-/* NOTE: the seccomp syscall whitelist is currently disabled for testing
+/* NOTE: the seccomp syscall allowlist is currently disabled for testing
  *       purposes, but unless we can verify all of the supported ABIs before
- *       our next release we may have to enable the whitelist */
+ *       our next release we may have to enable the allowlist */
 #define SYSCALL_WHITELIST_ENABLE	0
```
drakenclimber, line 41: Do you want to change the name of this #define as well?

pcmoore: /me smacks forehead

Yes, that needs to be changed as well.  Can you tell I didn't pass the '-i' argument to `grep`? :)

pcmoore:
Merged via 0e762521d604612bb4dca8867d4a428a5e6cae54, closing this out.

src/system.c
drakenclimber: :)

PR #247, opened 2020-06-11
RFE: rename hash() to seccomp_hash() to resolve name collisions when static linking

abudnik:
Since hash() is exposed as a strong symbol, it may lead to link-time
errors when another library defines the same function called `hash()`.
This patch fixes the issue by adding a `seccomp_` prefix.

Testing done: `make check`

### DIFF
--- a/src/gen_bpf.c
+++ b/src/gen_bpf.c
@@ -578,10 +578,10 @@ static int _hsh_add(struct bpf_state *state, struct bpf_blk **blk_p,
 		return -ENOMEM;
 
 	/* generate the hash */
-	h_val_tmp[0] = hash(blk->blks, _BLK_MSZE(blk));
-	h_val_tmp[1] = hash(&blk->acc_start, sizeof(blk->acc_start));
-	h_val_tmp[2] = hash(&blk->acc_end, sizeof(blk->acc_end));
-	h_val = hash(h_val_tmp, sizeof(h_val_tmp));
+	h_val_tmp[0] = seccomp_hash(blk->blks, _BLK_MSZE(blk));
+	h_val_tmp[1] = seccomp_hash(&blk->acc_start, sizeof(blk->acc_start));
+	h_val_tmp[2] = seccomp_hash(&blk->acc_end, sizeof(blk->acc_end));
+	h_val = seccomp_hash(h_val_tmp, sizeof(h_val_tmp));
 	blk->hash = h_val;
 	blk->flag_hash = true;
 	blk->node = NULL;
--- a/src/hash.c
+++ b/src/hash.c
@@ -38,7 +38,7 @@ static inline uint32_t fmix32(uint32_t h)
 }
 
 /* NOTE: this is an implementation of MurmurHash3_x86_32 */
-uint32_t hash(const void *key, size_t length)
+uint32_t seccomp_hash(const void *key, size_t length)
 {
 	const uint8_t *data = (const uint8_t *)key;
 	const uint32_t *blocks;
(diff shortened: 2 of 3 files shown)


pcmoore:
Hi @abudnik.

Please take a look at the [CONTRIBUTING.md](https://github.com/seccomp/libseccomp/blob/master/CONTRIBUTING.md) file, it helps describe how to properly submit patches to the project.  While you tested your changes (thank you!) you didn't include a sign-off line which unfortunately blocks us from accepting your changes.

Also, can you describe the problem you are having, and perhaps provide a simple reproducer if you can?  We currently mark all of our symbols with a hidden visibility and only expose the public API with a default visibility; my understanding is that this should protect against collisions between the internal libseccomp functions and functions defined by the caller.  Is this not correct?  Are you linking statically to libseccomp?

abudnik:
The error message is:
```
/usr/bin/ld: error: ../3rdparty/zookeeper-3.4.8/src/c/.libs/libzookeeper_mt.a(hashtable.o): multiple definition of 'hash'
/usr/bin/ld: ../3rdparty/libseccomp-2.4.3/src/.libs/libseccomp.a(libseccomp_la-hash.o): previous definition here
collect2: error: ld returned 1 exit status
```
We link statically against libseccomp _only_ to run tests in our internal CI. In all other cases, we link dynamically to respect license compatibility.

```
$ nm 3rdparty/libseccomp-2.4.3/src/.libs/libseccomp.a | grep hash

libseccomp_la-gen_bpf.o:
                 U hash
libseccomp_la-hash.o:
0000000000000000 T hash
```

pcmoore:
Okay, this makes sense now; it's a static build problem, not a shared library problem.

This is a tricky problem, as renaming just `hash(...)` leaves the door open for potential problems with other symbols.  While most of our other symbols are prefixed with some sort of namespace, they are not necessarily libseccomp prefixes which means they could still be problematic.

I only spent a little bit of time searching for a more general solution to this, but the results were not encouraging.  If someone has an idea that doesn't involve large renaming patches I would be interested in hearing about your idea.

I'm going to move this off the v2.5.0 release and adjust the priority down until we sort this out.

pcmoore:
Having thought about this some more, I'm not sure this is something we want to merge.  Any thoughts @drakenclimber?

While yes, we could rename this one function, but surely we would have the same problem with another function at some point.  If we rename one internal function we should really rename them all, and that is where things start to get a bit silly.  Compounding things is that this is only a problem when the libseccomp sources are linked statically with the application; this is far from the common case these days and I'm not sure we want to add a lot of code churn just for this, very limited use case.

drakenclimber:
> Having thought about this some more, I'm not sure this is something we want to merge. Any thoughts @drakenclimber?
> 
> While yes, we could rename this one function, but surely we would have the same problem with another function at some point. If we rename one internal function we should really rename them all, and that is where things start to get a bit silly. Compounding things is that this is only a problem when the libseccomp sources are linked statically with the application; this is far from the common case these days and I'm not sure we want to add a lot of code churn just for this, very limited use case.

Hmmm...

I admit that I have been bitten in the past by similar problems to this.  (And, yes, statically linking usually isn't the right answer, but the decision had been made long before I started touching that particular code.)  So I do have some sympathy for you there :).

I agree that I don't want to deal with recurring issues like this, as that would be a lot of code churn and, more importantly, time.

With all that said, I'm cool with letting this one exception slide through. :)

pcmoore:
Okay, I'm going to close this out as WONTFIX.  I apologize that this isn't likely the desired outcome @abudnik, but I don't think supporting static linking is something we can reliably support moving forward.

Regardless, thank you for the patch/PR and your patience.

Issue #240, opened 2020-05-28
BUG: gen_bpf_generate() doesn't properly handle failure


Xyene:
Hi,

First off, thanks for libseccomp -- we've been happily using it in production for several years now, and haven't run into any issues (until now). I'm not sure if this is a bug in our code, a misunderstanding of the documentation, or something else -- but I've spent the past month trying to track this down to no avail.

We recently upgraded packages in our Docker containers, which included an upgrade from libseccomp 2.3.3 (version in Debian stable repos) to 2.4.3. There were other system packages that also got upgraded, but I did not record them. Our kernel was not upgraded, and is version 4.19.0-8-amd64.

We use `SCMP_ACT_TRACE`, and build a filter consisting only of `SCMP_ACT_ALLOW` rules that are added using native syscall numbers, rather than libseccomp's pseudo-numbers. We fork off a 64-bit helper process that builds and loads the seccomp filter before `exec`-ing another 64-bit binary.

For reference, [this](https://github.com/DMOJ/judge-server/blob/ae73a48defdda19a54a9997db07014eb9f274962/dmoj/cptbox/helper.cpp#L156-L189) is the entirety of our seccomp initialization routine, using similar error-checking to the [`seccomp_rule_add` man page](https://www.man7.org/linux/man-pages/man3/seccomp_rule_add.3.html).

However, our call to `seccomp_load` has started returning `-EINVAL`, on the order of magnitude of 1 / 100,000 process initializations. (Not being able to reliably reproduce it has made this tedious to debug.) There were no code changes to our application in during this time. The syscalls added to the filter are identical across all runs.

Any ideas on what could be going wrong (or even how to dig further into what's going wrong), or if this is expected in some way? There aren't a whole lot of dynamic moving parts, and I couldn't find anything in the documentation about why this might be happening.

pcmoore:
Hi @Xyene,

There aren't a lot of places that return -EINVAL in the seccomp_load() code path.  Based on a quick examination of the libseccomp v2.4.3 code it looks like it is either due to an invalid `scmp_filter_ctx` or the kernel complaining about the `prctl(...)` call which loads the filter.

Considering that v2.4.3 generally works, and you haven't changed your kernel, it seems doubtful that the `prctl(...)` call is the cause which leads us to an invalid filter context.  Have you noticed any other odd behavior in your program since the upgrade?  I'm wondering if there is a memory corruption problem elsewhere that is causing the problem.

[...]

PR #256, opened 2020-06-16
BUG: don't return a bogus pointer when gen_bpf_generate() fails

pcmoore:
In the case where gen_bpf_generate() fails due to a
_gen_bpf_build_bpf() failure we were returning a pointer to a BPF
program block which we had already freed - oops.  Thankfully the
fix is trivial.

Special thanks to Tudor Brindus for not only reporting the problem
but providing a very detailed root cause analysis.

Fixes: #240

### DIFF
--- a/src/gen_bpf.c
+++ b/src/gen_bpf.c
@@ -1968,6 +1968,8 @@ struct bpf_program *gen_bpf_generate(const struct db_filter_col *col)
 	rc = _gen_bpf_build_bpf(&state, col);
 	if (rc == 0)
 		state.bpf = NULL;
+	else
+		prgm = NULL;
 	_state_release(&state);
 
 	return prgm;


pcmoore:
I'm fairly certain this will squash your bug @Xyene, but if you have the chance it would be nice if you could try the patch just to be sure.

Xyene:
Thanks for the quick fix! This looks appropriate from my understanding of the issue as well.

I actually switched the VMs to use libseccomp built from HEAD after finding it was fixed there, but it looks like the stars are no longer aligned and what used to be a solid reproduction method no longer triggers (regardless of whether 2.4.3 or HEAD is used)...

I think I've fixed the error on my part causing segfaults to not be reported properly, and have restarted testing. My remaining concern regarding #240 is that while not having a double-free in case of `_gen_bpf_build_bpf` failure is good, it doesn't explain why `_gen_bpf_build_bpf` would fail in the first place -- which would still bubble back up to `seccomp_load` even with this fix :slightly_smiling_face: 

Hopefully I'll have more info soon. I'm starting to suspect something may be up with jemalloc given there are no errors under valgrind or with glibc, but we'll see.

drakenclimber:
Thanks so much for the help @Xyene!

pcmoore:
Merged via c0a6e6fd15f74c429a0b74e0dfd4de5a29aabebd, thanks everyone!

PR #257, opened 2020-06-17
RFE: propagate errors from _gen_bpf_build_bpf helpers

Xyene:
Prior to this commit, `_gen_bpf_build_bpf` would mask some errors that
occurred in helper functions as `EFAULT`, even if they were not `EFAULT` to
begin with.

Refs #240.

### DIFF
--- a/src/gen_bpf.c
+++ b/src/gen_bpf.c
@@ -2144,6 +2144,7 @@ static int _gen_bpf_build_bpf(struct bpf_state *state,
 				break;
 			default:
 				/* fatal error */
+				rc = -EFAULT;
 				goto build_bpf_free_blks;
 			}
 			switch (i_iter->jf.type) {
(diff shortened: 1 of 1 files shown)


pcmoore:
A quick note for the future @Xyene, I'm a big fan of making the git commit descriptions self-contained.  The basic idea is that someone working off-line, or in a post-GH world, should be able to look through the git log and make sense of the changes.

Unfortunately this means GH shorthand, e.g. "Refs: #240", aren't very helpful.  You can still add them, but the links shouldn't replace adding text to the commit description and if you are going to provide a GH reference, it should be the full URL and not just a "#<issue>".

Regardless, I really appreciate all the help you've provided - thank you!  I'm reviewing this patch now ...

pcmoore:
This looks good to me so I've merged it via 12cf0074be3fa87ef33eecde6848176f03cd1460, thanks again @Xyene.

@drakenclimber this looked pretty straightforward to me so I went ahead and merged it (obviously), if you've got concerns let me know and we can either do a fixup patch or revert this.

drakenclimber:
> @drakenclimber this looked pretty straightforward to me so I went ahead and merged it (obviously), if you've got concerns let me know and we can either do a fixup patch or revert this.

No objections here.  Thanks @pcmoore and @Xyene

Issue #259, opened 2020-06-25
RFE: allow adding an exact syscall number for an architecture


Xyene:
Hi,

I have a usage of libseccomp that involves a 64-bit process setting up a filter for and then launching a 32-bit process, with `SCMP_ACT_TRACE` and using `ptrace` for syscall emulation.

The code already has to maintain a syscall name to per-arch ID (similar to libseccomp's [syscall.csv](https://github.com/seccomp/libseccomp/blob/0a6c5f50080691a7984fc12c8b9d605819f66ad5/src/syscalls.csv) in spirit), since it falls back to pure `ptrace` if `seccomp` is not available. This means that in order to use libseccomp's handy `SCMP_SYS` helpers, we'd need to maintain *another* mapping to use only on the seccomp path. Possible, but icky.

Reading the docs, I would have expected the following code to set up a filter allowing `fork` for x86, and cause traps for all other syscalls (most error-checking elided for brevity):

```c
#include <stdio.h>
#include <string.h>
#include <seccomp.h>

void pfc(scmp_filter_ctx ctx) {
    seccomp_export_pfc(ctx, 2);
    fprintf(stderr, "-----\n");
}

int main(int argc, char **argv) {
    scmp_filter_ctx ctx = seccomp_init(SCMP_ACT_TRACE(0));
    seccomp_arch_add(ctx, SCMP_ARCH_X86);
    seccomp_arch_remove(ctx, SCMP_ARCH_NATIVE);
    pfc(ctx);

    int rc;
    // fork is syscall 2 on x86
    if ((rc = seccomp_rule_add_exact(ctx, SCMP_ACT_ALLOW, 2, 0))) {
        fprintf(stderr, "seccomp_rule_add: %s\n", strerror(-rc));
    }

    pfc(ctx);
    return 0;
}
```

Compiling and running this example produces:

```
$ gcc -o test test.c -lseccomp                                                
$ file test                                                                 
test: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=85c3d0a91296a9107f390c1124e69f1d45387134, for GNU/Linux 3.2.0, not stripped
$ ./test
#
# pseudo filter code start
#
# filter for arch x86 (1073741827)
if ($arch == 1073741827)
  # default action
  action TRACE(0);
# invalid architecture action
action KILL;
#
# pseudo filter code end
#
-----
#
[...]

PR #261, opened 2020-06-25
RFE: clarify seccomp_rule_add syscall translation behavior

Xyene:
Refs #259.

### DIFF
--- a/doc/man/man3/seccomp_rule_add.3
+++ b/doc/man/man3/seccomp_rule_add.3
@@ -147,7 +147,10 @@ value directly using the standard
 values, in order to ensure proper operation across multiple architectures it
 is highly recommended to use the
 .BR SCMP_SYS ()
-macro instead.  See the EXAMPLES section below.
+macro instead.  See the EXAMPLES section below.  It is also important to
+remember that regardless of the architectures present in the filter, the
+syscall numbers used in filter rules are interpreted in the context of the
+native architecture.
 .P
 Starting with Linux v4.8, there may be a need to create a rule with a syscall
 value of -1 to allow tracing programs to skip a syscall invocation; in order


doc/man/man3/seccomp_rule_add.3
```diff
@@ -147,7 +147,11 @@ value directly using the standard
 values, in order to ensure proper operation across multiple architectures it
 is highly recommended to use the
 .BR SCMP_SYS ()
-macro instead.  See the EXAMPLES section below.
+macro instead.  See the EXAMPLES section below.  If using
+.B __NR_syscall
+values in a context containing non-native architectures, the syscall
+corresponding to the value in the native architecture will be used in all
+non-native architectures.
```
pcmoore, line 154: I not sure that reads quite right, how about the following?

> ... it is highly recommended to use the SCMP_SYS() macro instead, see the EXAMPLES section below.  It is also important to remember that regardless of the architectures present in the filter, the syscall numbers passed to the libseccomp rule add functions should always be in the context of the native architecture.

pcmoore: @Xyene what did you think of the substitute text above?

Xyene: Sorry about the delay, I got DOS'd by real-life stuff -- I've committed a slightly reworded variant of your suggested text, emphasizing that libseccomp is performing some "interpreting" of the given syscall value.

pcmoore:
Thanks for getting back to this, I definitely understand competing priorities and I can assure you we appreciate the help regardless of where libseccomp falls in your personal priority list :)

This  updated text looks good to me, any objections @drakenclimber?

drakenclimber:
Looks good to me :+1: 

pcmoore:
Merged via fa6264bd9d674d5bd141d1153ed1d77e654b4af1, thanks!

PR #263, opened 2020-06-30
RFE: update the syscall table for Linux v5.8.0-rc5

pcmoore:
This is a work-in-progress PR to track the syscall table changes leading up to the libseccomp v2.5.0 release.  It will be updated as the Linux syscall table changes.


pcmoore:
Updated to 5.8-rc5.

Issue #252, opened 2020-06-15
BUG: don't generate BPF for pseudo-syscalls


pcmoore:
See issue #249, specifically [this comment](https://github.com/seccomp/libseccomp/issues/249#issuecomment-644111675):

> > Why check for non-existent syscall 4294957285 in actually generated bpf?
>
> We shouldn't, and we didn't used to do that, but it looks like current libseccomp has a bug here.

Credit to @vt-alt for reporting this bug.

pcmoore:
Trivial reproducer:
```c
#include <stdlib.h>
#include <errno.h>

#include <seccomp.h>

#include "util.h"

int main(int argc, char *argv[])
{
 int rc;
 struct util_options opts;
 scmp_filter_ctx ctx = NULL;

 rc = util_getopt(argc, argv, &opts);
 if (rc < 0)
        goto out;

 ctx = seccomp_init(SCMP_ACT_KILL);
 if (ctx == NULL)
        return ENOMEM;

 rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW,
                      SCMP_SYS(arm_fadvise64_64), 0);
 if (rc < 0)
        goto out;

 rc = util_filter_output(&opts, ctx);
 if (rc)
        goto out;

out:
 seccomp_release(ctx);
 return (rc < 0 ? -rc : rc);
}
```

pcmoore:
To be clear, we should still emit the pseudo-syscalls in the PFC, for debugging purposes, but we shouldn't emit a (useless) BPF filter rule.

vt-alt:
Thanks for creating the issue from my report.

> To be clear, we should still emit the pseudo-syscalls in the PFC, for debugging purposes, but we shouldn't emit a (useless) BPF filter rule.

Please rethink this. I believe this will only complicate and obscure things if PFC will not reflect BPF.

pcmoore:
> Please rethink this. I believe this will only complicate and obscure things if PFC will not reflect BPF.

If the syscall is missing from the PFC we will receive a number of bogus bug reports talking about how the library is failing to add a filter for (non-existent) syscalls.

For those who are understand the concept of the pseudo-syscalls it is a trivial exercise to remove them from the PFC output.  It's also worth mentioning that the PFC output is not intended to be an exact copy of the BPF output, it's there simply as a debugging tool and an easy way to visualize the generated filter code.

vt-alt:
> easy way to visualize the generated filter code

[...]

PR #264, opened 2020-07-02
BUG: Don't generate BPF for psuedo-syscalls

drakenclimber:
This patchset:
- Removes pseudo-syscalls from the output BPF filter
- Fixes a couple potential bugs in the x32 and arm code where we were doing subtraction/addition rather than bitwise operations
- Updates two tests to work with the BPF/pseudo-syscalls change


drakenclimber:
Fixes Issue #252

src/gen_bpf.c
```diff
@@ -1291,6 +1291,50 @@ static int _gen_bpf_insert(struct bpf_state *state, struct bpf_instr *instr,
 	return rc;
 }
 
+/**
+ * Omit syscall from the BPF filter?
```
pcmoore, line 1295: Big time nitpick, but typically the one-liner summaries are statements not questions, e.g. "Decide if we need to omit the syscall from the BPF filter".

src/gen_bpf.c
```diff
@@ -1532,11 +1576,17 @@ static int _gen_bpf_syscalls(struct bpf_state *state,
...
-					   bintree_levels, db->syscall_cnt,
+					   bintree_levels, syscall_cnt,
 					   &empty_cnt);
 		if (rc < 0)
 			goto out;
+
+		/* reset the syscall_cnt variable because later in this
+		 * function it's used as a counter
+		 */
+		syscall_cnt = 0;
```
pcmoore, line 1589: We can drop this special handling if we move the `syscall_cnt` initialization to 0 from the declaration line to right about the for-loop where it is first used.

src/gen_bpf.c
```diff
@@ -1555,6 +1605,8 @@ static int _gen_bpf_syscalls(struct bpf_state *state,
 	for (s_iter = s_tail; s_iter != NULL; s_iter = s_iter->pri_prv) {
 		if (!s_iter->valid)
 			continue;
+		if (_skip_syscall(state, s_iter))
```
pcmoore, line 1608: Same as above, might as well combine this with the valid check.

src/gen_bpf.c
drakenclimber: :) Can you tell I was in a Java frame of mind?  Thanks, I like your wording better.

src/gen_bpf.c
```diff
@@ -1291,6 +1291,50 @@ static int _gen_bpf_insert(struct bpf_state *state, struct bpf_instr *instr,
...
+ * @param syscall syscall being tested
+ * @return true if syscall is to be skipped, false otherwise
+ */
+static inline bool _skip_syscall(struct bpf_state *state,
+				 struct db_sys_list *syscall)
+{
+	/* psuedo-syscalls should not be added to the filter unless explicity
+	 * requested via SCMP_FLTATR_API_TSKIP
+	 */
+	if ((syscall->num & 0x80000000) &&
```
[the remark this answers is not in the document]
drakenclimber: I'm glad you brought this up.  I copy-pasta'ed this from the proposed patchset you wrote [here](https://github.com/seccomp/libseccomp/issues/252#issuecomment-644154303), and I was curious why.  I'll definitely switch to using `<` as that reads so much better.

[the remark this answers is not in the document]
drakenclimber: Looks like you used the bitwise operator because num is defined as an unsigned int.  I'll use `<` and a cast to an int.

pcmoore:
Merged at HEAD 2f3ae1fe7eaa1195e29979ae341bfc288bc69ee5, thanks!

vt-alt:
Thanks!

Issue #245, opened 2020-06-06
BUG: the seccomp_attr_get/set(3) manpages do not document SCMP_FLTATR_CTL_OPTIMIZE


pcmoore:
Not much to say beyond the subject line.  @drakenclimber do you want to take this one?

drakenclimber:
Sure

pcmoore:
Ping - I just wanted to make sure you didn't forget about this one @drakenclimber :)

drakenclimber:
Yep, I should be able to get to it this week.  But thanks for the reminder; I can use all the help I can get these days :)

pcmoore:
Closed via PR #266.

PR #266, opened 2020-07-14
BUG: the seccomp_attr_get/set(3) manpages do not document SCMP_FLTATR_CTL_OPTIMIZE

drakenclimber:
This commit documents the usage of SCMP_FLTATR_CTL_OPTIMIZE in the seccomp_attr_set.3 man page.

Addresses Issue #245

### DIFF
--- a/doc/man/man3/seccomp_attr_set.3
+++ b/doc/man/man3/seccomp_attr_set.3
@@ -101,6 +101,32 @@ Defaults to off (
 .I value
 == 0).
 .TP
+.B SCMP_FLTATR_CTL_OPTIMIZE
+A flag to specify the optimization level of the seccomp filter.  By default
+libseccomp generates a set of sequential \'if\' statements for each rule in
+the filter.
+.BR seccomp_syscall_priority(3)
+can be used to prioritize the order for the default cause.  The binary tree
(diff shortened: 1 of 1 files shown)


drakenclimber:
Oh, and if you hate reading man markup, here's what it looks like:

```
       SCMP_FLTATR_CTL_OPTIMIZE
              A  flag  to  specify the optimization level of the seccomp filter.  By default libseccomp generates a
              set of sequential 'if' statements for each rule in the filter.   seccomp_syscall_priority(3)  can  be
              used  to  prioritize  the order for the default cause.  The binary tree optimization sorts by syscall
              numbers and generates consistent O(log n) filter traversal for every rule in the filter.  The  binary
              tree  may  be  advantageous for large filters.  Note that seccomp_syscall_priority(3) is ignored when
              SCMP_FLTATR_CTL_OPTIMIZE == 2.

              The different optimization levels are described below:

              0      Reserved value, not currently used.

              1      Rules sorted by priority and complexity (DEFAULT).

              2      Binary tree sorted by syscall number.
```

pcmoore:
Looks good to me, merged via 43d63abc1ef2ae3e9386c4b1733a774575d144d8, thanks!

PR #268, opened 2020-07-17
tests: Explicitly specify the archs in the sim binary tree test

drakenclimber:
Test 53-sim-binary_tree removes the native architecture and
builds the binary tree on aarch64, ppc64le, and x86_64.  But the
tests file was testing for "all" architectures which led to test
failures on other systems like s390x.  This commit replaces the
"all" arch with only the architectures in the test.


pcmoore:
Merged via a6de31f0d90f00bfd6900480857f6c71f1d3c8ef and 919be9f8fe62dd93edda00ba5fac7eaf11ff1e1f, thanks!

PR #269, opened 2020-07-19
RFE: add support for openat2 to the release-2.4 branch

intelfx:
I'm not sure what exactly is the purpose of `include/seccomp-syscalls.h` and what are the `__PNR_`/`__SNR_` prefixes for, so I may have gotten that wrong.

Tests pass on x86_64. Not sure if new tests are needed.

### DIFF
--- a/include/seccomp-syscalls.h
+++ b/include/seccomp-syscalls.h
@@ -273,6 +273,7 @@
 #define __PNR_timerfd_settime64			-10239
 #define __PNR_utimensat_time64			-10240
 #define __PNR_ppoll				-10241
+#define __PNR_openat2				-10242
 
 /*
  * libseccomp syscall definitions
@@ -1285,6 +1286,12 @@
 #endif
 
 #define __SNR_openat			__NR_openat
+
+#ifdef __NR_openat2
+#define __SNR_openat2			__NR_openat2
+#else
+#define __SNR_openat2			__PNR_openat2
+#endif
 
 #ifdef __NR_pause
 #define __SNR_pause			__NR_pause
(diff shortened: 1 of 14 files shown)


include/seccomp-syscalls.h
```diff
@@ -273,6 +273,7 @@
 #define __PNR_timerfd_settime64			-10239
 #define __PNR_utimensat_time64			-10240
 #define __PNR_ppoll				-10241
+#define __PNR_openat2				-10242
```
pcmoore, line 276: A `__PNR_xxx` define isn't needed as `openat2` is defined for every ABI we currently support.

include/seccomp-syscalls.h
```diff
@@ -1286,6 +1287,12 @@
 
 #define __SNR_openat			__NR_openat
 
+#ifdef __NR_openat2
+#define __SNR_openat2			__NR_openat2
+#else
+#define __SNR_openat2			__PNR_openat2
+#endif
```
pcmoore, line 1294: See my comment above, this can be simplified since we don't need `__PNR_openat2`.

pcmoore:
> @pcmoore - I noticed that the release-2.4 arch-syscall-validate doesn't validate parisc. Do you think we should backport those patches?

That's one of those judgement call decisions that I can see making sense either way.  On one hand it would be nice to have that extra verification, but on the other hand the release-2.4 branch is effectively in "long term support" and there is an argument to be made about not rocking the boat.  As you are managing the "stable" branches/releases, I'll let you cast the tie breaking vote :)

pcmoore:
Hi @intelfx, any chance you could do a repin to address the feedback above?

drakenclimber:
Since I was updating many other syscalls, I also added `openat2` support.

Fixed via commit https://github.com/seccomp/libseccomp/commit/b3206ad5645dceda89538ea8acc984078ab697ab

Issue #274, opened 2020-07-26
BUG: compile error on MIPS architecture


pudh4418:
Hi,

Compiling libseccomp v2.5.0 on MIPS will bring errors like this:
```
libtool: compile:  mips64el-unknown-linux-gnuabi64-gcc -DHAVE_CONFIG_H -I. -I.. -I../include -I../include -Wall -g -O2 -fPIC -DPIC -fvisibility=hidden -g -O2 -c arch-x86.c  -fPIC -DPIC -o .libs/libseccomp_la-arch-x86.o
syscalls.h:44:6: error: expected identifier or '(' before numeric constant
   44 |  int mips;
      |      ^~~~
```

This error will also appear when cross compiling MIPS lib from x86-64. A preliminary analysis shows that GCC defines `mips` for MIPS architecture, thus the code becomes `int 1;` .
Maybe we can change it to a different name?

Thanks.

pcmoore:
Thanks for the report @pudh4418, I'll take a look.

pcmoore:
Hi @pudh4418, can you check the commit below?  I think it should fix your problem:
* https://github.com/pcmoore/misc-libseccomp/commit/a445a16c4734fff30910178f415a475791a5f267

Also, I'd like to give you credit for reporting the bug in the commit via the `Reported-by:` tag, but in order to do that I need your name and email.  If you don't want to share the information, that's fine too, I'll leave the commit as it is now.

pudh4418:
It seems the fix is not that easy. After applying your patch, error messages turn to:
```
libtool: compile:  mips64el-unknown-linux-gnuabi64-gcc -DHAVE_CONFIG_H -I. -I.. -I../include -I../include -Wall -g -O2 -fPIC -DPIC -fvisibility=hidden -g -O2 -c arch-s390x.c  -fPIC -DPIC -o libseccomp_la-arch-s390x.o >/dev/null 2>&1
In file included from arch.h:26,
                 from syscalls.c:25:
syscalls.c: In function 'mips_syscall_resolve_name':
syscalls.h:55:63: error: 'struct arch_syscall_table' has no member named 'arch_1'; did you mean 'arch_x86'?
   55 | #define OFFSET_ARCH(NAME) offsetof(struct arch_syscall_table, arch_##NAME)
      |                                                               ^~~~~
syscalls.c:31:37: note: in expansion of macro 'OFFSET_ARCH'
[...]

PR #279, opened 2020-08-02
BUG: undefine "mips" to prevent build problems for MIPS targets

pcmoore:
It turns out that the MIPS GCC compiler defines a "mips" cpp macro
which was resulting in build failures on MIPS so we need to
undefine the "mips" macro during build.  As this should be safe
to do in all architectures, just add it to the compiler flags by
default.

This was reported in the following GH issue:
* https://github.com/seccomp/libseccomp/issues/274

Reported-by: Rongwei Zhang <pudh4418@gmail.com>
Suggested-by: Rongwei Zhang <pudh4418@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -65,9 +65,11 @@ m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 
 dnl ####
 dnl build flags
+dnl NOTE: the '-Umips' is here because MIPS GCC compilers "helpfully" define it
+dnl       for us which wreaks havoc on the build
 dnl ####
 AM_CPPFLAGS="-I\${top_srcdir}/include -I\${top_builddir}/include"
-AM_CFLAGS="-Wall"
+AM_CFLAGS="-Wall -Umips"
 AM_LDFLAGS="-Wl,-z -Wl,relro"
 AC_SUBST([AM_CPPFLAGS])
 AC_SUBST([AM_CFLAGS])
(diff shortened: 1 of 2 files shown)


pcmoore:
Assuming this looks okay and is merge into the master branch, I'll also merge this into the release-2.5 branch.

pcmoore:
@pudh4418 would you mind verifying that this works for you?

drakenclimber:
> ```diff
> arch_syscall_check_CFLAGS = ${AM_CFLAGS} ${CODE_COVERAGE_CFLAGS}
> ```

Good catch.  I looked through the autogenerated `Makefile` and `AM_CFLAGS` wasn't propagated to `COMPILE`, `LTCOMPILE`, and `LINK` without the above change.

drakenclimber:
What about the various tools like `scmp_bpf_disasm`?  Do they need to have `AM_FLAGS` propagated as well?

pcmoore:
Thanks @pudh4418 for catching the problem in "src/Makefile.am", I just updated the PR with your suggested fix.  Since @drakenclimber already ACK'd the suggestion I'll push it to the master branch as soon as the CI checks come back clean.

pcmoore:
> What about the various tools like scmp_bpf_disasm? Do they need to have AM_FLAGS propagated as well?

I don't believe this should be a concern as there are no sub-directories under "tools/".  I believe the `AM_FLAGS` variable is used by the parent directory, this case the base directory, when calling `make` on child directories.  That said, I am definitely not an autotools expert so if someone knows better than I please correct me.

However, regardless of what we do with `AM_FLAGS`, I think that should be addressed in a separate issue/PR as I don't believe that impacts the MIPS build breakage that we are trying to address here.

Issue #282, opened 2020-08-08
BUG: pseudo-syscalls not working correctly on x32


hvdijk:
Commit 9b129c41ac1f43d373742697aa2faf6040b9dfab introduces the concept of "syscall munging" to subtract some base number from syscall numbers to make it easier to use them as an offset into a table. Despite messages in this commit specifically addressing negative pseudo syscall numbers, this does not actually handle negative pseudo syscall numbers properly, which shows up as a test failure in 29-sim-pseudo_syscall for at least x32 and probably all architectures that use munging.

29-sim-pseudo_syscall does `rc = seccomp_rule_add(ctx, SCMP_ACT_KILL, SCMP_SYS(sysmips), 0);`, which passes `__PNR_sysmips` as `syscall`. This syscall remains unmodified until `_db_col_rule_add` calls `arch_filter_rule_add`, which calls `arch_syscall_translate`, which calls `arch_syscall_resolve_num`. For x32, this resolves to a call to `x32_syscall_resolve_num_munge`, which unconditionally subtracts `X32_SYSCALL_BIT` and then calls `x32_syscall_resolve_num`. However, `x32_syscall_resolve_num` is only set up to handle `__PNR_sysmips`, not `__PNR_sysmips - X32_SYSCALL_BIT`! In order to get this test to pass after this commit, it would somehow need to know that it needs to add `X32_SYSCALL_BIT` to the syscall number, but that it unreasonable, the whole point of these pseudo-syscalls is to be able to write the same code across all architectures regardless of which syscalls are actually supported.

If syscall numbers can never be negative for any architecture, it is possible to fix this problem by changing all the munging/unmunging routines to leave negative numbers alone: they must be pseudo syscall numbers. This works for x32, but I do not have enough information to say whether this is safe for all architectures. (Edit: it would also require some special handling to prevent an invalid syscall number from becoming a valid pseudo-syscall number by the subtraction of `X32_SYSCALL_BIT`. Additionally, after 9b129c41ac1f43d373742697aa2faf6040b9dfab, but before v2.5.0, the subtraction has been changed to a bitwise operation which has the same problem: that too can inadvertently treat invalid syscall numbers as valid ones. That should have been kept doing addition and subtraction.)

pcmoore:
Hi @hvdijk, do you have a reproducer that you can share?

hvdijk:
The reproducer is already in the test suite: just do an x32 build and run the test suite's 29-sim-pseudo_syscall test.

[...]

PR #283, opened 2020-08-08
BUG: store syscall numbers directly

hvdijk:
This removes the need to munge and unmunge numbers and ensures that
seccomp_rule_add can handle pseudo-syscall numbers again.

Fixes: https://github.com/seccomp/libseccomp/issues/282

Signed-off-by: Harald van Dijk <harald@gigawatt.nl>

### DIFF
--- a/src/arch-arm.c
+++ b/src/arch-arm.c
@@ -26,58 +26,13 @@
 #include "arch.h"
 #include "arch-arm.h"
 
-#define __SCMP_NR_OABI_SYSCALL_BASE     0x900000
-#define __SCMP_ARM_NR_BASE              0x0f0000
-
-/* NOTE: we currently only support the ARM EABI, more info at the URL below:
- *       -> http://wiki.embeddedarm.com/wiki/EABI_vs_OABI */
-#if 1
-#define __SCMP_NR_BASE                  0
-#else
-#define __SCMP_NR_BASE                  __SCMP_NR_OABI_SYSCALL_BASE
-#endif
-
-/**
- * Resolve a syscall name to a number
- * @param name the syscall name
- *
- * Resolve the given syscall name to the syscall number using the syscall table.
- * Returns the syscall number on success, including negative pseudo syscall
- * numbers; returns __NR_SCMP_ERROR on failure.
- *
- */
-int arm_syscall_resolve_name_munge(const char *name)
-{
-	int sys;
-
-	sys = arm_syscall_resolve_name(name);
-	if (sys == __NR_SCMP_ERROR)
-		return sys;
-
-	return (sys | __SCMP_NR_BASE);
-}
-
-/**
- * Resolve a syscall number to a name
- * @param num the syscall number
- *
- * Resolve the given syscall number to the syscall name using the syscall table.
- * Returns a pointer to the syscall name string on success, including pseudo
- * syscall names; returns NULL on failure.
- *
- */
-const char *arm_syscall_resolve_num_munge(int num)
-{
-	return arm_syscall_resolve_num(num & (~__SCMP_NR_BASE));
-}
-
 const struct arch_def arch_def_arm = {
 	.token = SCMP_ARCH_ARM,
 	.token_bpf = AUDIT_ARCH_ARM,
 	.size = ARCH_SIZE_32,
 	.endian = ARCH_ENDIAN_LITTLE,
-	.syscall_resolve_name = arm_syscall_resolve_name_munge,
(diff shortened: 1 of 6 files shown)


pcmoore:
Hi @hvdijk, thanks for the PR!  Unfortunately I have limited network access this week so I'm not sure I will be able to properly review this PR this week; I think @drakenclimber may also have some time conflicts limiting his ability to review.

That said, based on a quick inspection I don't think we want to completely remove the munging as has been done in this PR.  As a general rule, it would be good to find a solution to #282 that doesn't involve changing syscalls.csv.

hvdijk:
> That said, based on a quick inspection I don't think we want to completely remove the munging as has been done in this PR. As a general rule, it would be good to find a solution to #282 that doesn't involve changing syscalls.csv.

I'm not sure why syscalls.csv shouldn't just contain the numbers of syscalls, but okay, that should be easy to address by scripting those changes instead by modifying `src/arch-gperf-generate`. I'll try to update this PR to do that when I have some time.

pcmoore:
To be clear the bigger issue is my "I don't think we want to completely remove the munging as has been done in this PR" comment, not affecting the syscall.csv file is just a good indicator that something is being changed that we may not want to change.

hvdijk:
> To be clear the bigger issue is my "I don't think we want to completely remove the munging as has been done in this PR" comment, not affecting the syscall.csv file is just a good indicator that something is being changed that we may not want to change.

The existing syscall munging at the C level is fundamentally broken, it should not have been put in in the first place (especially without testing) and it should not be kept now. `seccomp_rule_add` is supposed to reject a syscall of 0 and accept a syscall of 0x40000000, but munging turns both of them into the same number (0), making it impossible to differentiate between the two.

I've updated the PR to modify arch-gperf-generate to handle the syscalls.csv transformation. It uses awk, which I hope is okay as it is already used in other places as well.

pcmoore:
> The existing syscall munging at the C level is fundamentally broken...

The general concept of syscall munging is an important part of libseccomp and isn't going to be removed any time soon. Perhaps we may need to adjust how it is used, or maybe even remove it for some of the arches mentioned here, but that is it.

As far as testing is concerned, libseccomp provides support for a number of niche arches which the maintainers of the library are not able to test themselves. We rely on others who have access to those arches to provide testing help; if you are able to do that @hvdijk it would be helpful.

hvdijk:
> > The existing syscall munging at the C level is fundamentally broken...
> 
> The general concept of syscall munging is an important part of libseccomp and isn't going to be removed any time soon. Perhaps we may need to adjust how it is used, or maybe even remove it for some of the arches mentioned here, but that is it.

Please actually look at the contents of the PR instead of making assumptions based on a comment. I meant the existing syscall munging at the C level for x32 and MIPS is fundamentally broken, and explained why. That is what I removed in this PR (and the no-op munging for ARM that would become equally broken if it got modified to not be a no-op), not all munging for all architectures. And that can hardly be such an important part of libseccomp if it's brand new in 2.5.0.

I do suspect some of the other munging that's being done for other architectures is not actually necessary but I did not make changes there.

> As far as testing is concerned, libseccomp provides support for a number of niche arches which the maintainers of the library are not able to test themselves. We rely on others who have access to those arches to provide testing help; if you are able to do that @hvdijk it would be helpful.

That's how I noticed the problem in the first place, by running the test suite for x32.

It's perfectly understandable that you don't run test suites for x32 and MIPS all the time, but when someone changes the way libseccomp works for x32 and MIPS in a way that doesn't impact other architectures, *that* really should have been tested on x32 or MIPS.

pcmoore:
@hvdijk at the moment I've got limited network access and I've been trying to do my best to provide you with timely and respectful feedback based on my limited resources. However, based on your increasingly disrespectful comments I'm growing less inclined to provide feedback right now.

Thank you for reporting the problem, we will look into both the original problem report, and your PR, when we have time.

hvdijk:
I don't see how "We rely on others who have access to those arches to provide testing help; if you are able to do that @hvdijk it would be helpful" could be read as "respectful feedback" rather than a passive aggressive snipe blaming me in my capacity as an x32 user for not having tested and reported this before the 2.5.0 release, considering I *am* providing testing help, and that's why my response to that was what it was.

Locally, I can just run libseccomp with this PR applied, so for me, this is not an urgent matter. If taking extra time before responding will allow for smoother communication, I think that's a great idea.

pcmoore:
Perhaps if *you* take the extra time before commenting and insulting those of us who work on putting together libseccomp things might go a bit smoother.

hvdijk:
PR withdrawn.

PR #287, opened 2020-08-16
BUG: use smaller "magic" numbers in 51-live-user_notification

pcmoore:
On an x32 based system, and perhaps other 32-bit systems, the magic
value in test 51-live-user_notification was too large and resulted
in a failed comparison even when the userspace notification mechanism
was working properly.

This patch addresses this problem by using the parent process's PID
for the magic number.  For all arches/ABIs we know it is a valid
return value for getpid() while also being one of the few PIDs that
the child process should never legitimately get from a call to
getpid().

This patch also restricts the use of SCMP_SYS() to only the
libseccomp API calls.  This should help us catch arches/ABIs that
don't implement getpid().  I'm also not sure we want to be
responsible for using SCMP_SYS() outside of the libseccomp API.

### DIFF
--- a/tests/51-live-user_notification.c
+++ b/tests/51-live-user_notification.c
@@ -21,6 +21,7 @@
 
 #include <sys/types.h>
 #include <sys/wait.h>
+#include <asm/unistd.h>
 #include <unistd.h>
 #include <seccomp.h>
 #include <signal.h>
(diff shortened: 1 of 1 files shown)


pcmoore:
Test results for x32 (x86_64 and others available via Travis):

```
% ../tools/scmp_arch_detect 
x32
% ../tools/scmp_api_level 
5
% ./regression -T live -b 51-live-user_notification
=============== Sun 16 Aug 2020 04:15:39 PM EDT ===============
Regression Test Report ("regression -T live -b 51-live-user_notification")
 batch name: 51-live-user_notification
 test mode:  c
 test type:  live
Test 51-live-user_notification%%001-00001 result:   SUCCESS
 batch name: 51-live-user_notification
 test mode:  python
 test type:  live
Test 51-live-user_notification%%001-00001 result:   SUCCESS
Regression Test Summary
 tests run: 2
 tests skipped: 0
 tests passed: 2
 tests failed: 0
 tests errored: 0
```

PR #288, opened 2020-08-16
BUG: ensure we don't "munge" pseudo syscall numbers

pcmoore:
A number of arches/ABIs have either syscall offsets (the MIPS
family) or specific bits (x32) which are applied to their normal
syscall numbers.  We generally handle that via "munging" in
libseccomp, and it works reasonably well.  Unfortunately we were
applying this munging process to the negative pseudo syscall
numbers as well and this was causing problems.

This patch fixes the various offset/bit arches/ABIs by not applying
the munging to the negative pseudo syscall numbers.

This resolves GH issue #284:
* https://github.com/seccomp/libseccomp/issues/284

Reported-by: Harald van Dijk <harald@gigawatt.nl>

### DIFF
--- a/src/arch-arm.c
+++ b/src/arch-arm.c
@@ -50,8 +50,9 @@ int arm_syscall_resolve_name_munge(const char *name)
 {
 	int sys;
 
+	/* NOTE: we don't want to modify the pseudo-syscall numbers */
 	sys = arm_syscall_resolve_name(name);
-	if (sys == __NR_SCMP_ERROR)
+	if (sys == __NR_SCMP_ERROR || sys < 0)
 		return sys;
 
 	return (sys | __SCMP_NR_BASE);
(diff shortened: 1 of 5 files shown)


pcmoore:
Testing results for x32 (x86_64 and others available via Travis CI):

```
% uname -r
5.4.48-gentoo-x86_64
% ./tools/scmp_arch_detect 
x32
% ./tools/scmp_api_level   
5
```

```
% ./regression -m c -m python
=============== Sun 16 Aug 2020 04:19:46 PM EDT ===============
Regression Test Report ("regression -m c -m python")
...
...
...
Regression Test Summary
 tests run: 5058
 tests skipped: 550
 tests passed: 5058
 tests failed: 0
 tests errored: 0
```

```
% ./regression -T live
=============== Sun 16 Aug 2020 04:23:15 PM EDT ===============
Regression Test Report ("regression -T live")
...
...
...
Regression Test Summary
 tests run: 20
 tests skipped: 2
 tests passed: 20
 tests failed: 0
 tests errored: 0
```

Notes on the live test run:
* The two skipped tests were due to the API level of "5", the skipped test was "58-live-tsync_notify" and is not arch/ABI specific so this shouldn't be an issue.
* The live test run also included the patch from PR #287 so that the live notification test worked properly on x32.

PR #290, opened 2020-08-18
BUG: skip the stat and open syscalls on riscv64 in tests 04 and 06

andreas-schwab:
Like aarch64, and all future architectures, riscv64 does not support the
open and stat syscalls.

### DIFF
--- a/tests/04-sim-multilevel_chains.c
+++ b/tests/04-sim-multilevel_chains.c
@@ -41,7 +41,7 @@ int main(int argc, char *argv[])
 	if (ctx == NULL)
 		return ENOMEM;
 
-	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(open), 0);
+	rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(openat), 0);
 	if (rc != 0)
 		goto out;
 
--- a/tests/04-sim-multilevel_chains.py
+++ b/tests/04-sim-multilevel_chains.py
@@ -30,7 +30,7 @@ from seccomp import *
 
 def test(args):
     f = SyscallFilter(KILL)
-    f.add_rule(ALLOW, "open")
+    f.add_rule(ALLOW, "openat")
     f.add_rule(ALLOW, "close")
     f.add_rule(ALLOW, "read",
                Arg(0, EQ, sys.stdin.fileno()),
--- a/tests/04-sim-multilevel_chains.tests
+++ b/tests/04-sim-multilevel_chains.tests
@@ -8,7 +8,7 @@
 test type: bpf-sim
 
 # Testname			Arch		Syscall		Arg0		Arg1		Arg2			Arg3	Arg4	Arg5	Result
-04-sim-multilevel_chains	all,-aarch64	open		0x856B008	4		N			N	N	N	ALLOW
+04-sim-multilevel_chains	all		openat		0		0x856B008	4			N	N	N	ALLOW
 04-sim-multilevel_chains	all		close		4		N		N			N	N	N	ALLOW
 04-sim-multilevel_chains	x86		read		0		0x856B008	0x7FFFFFFE		N	N	N	ALLOW
 04-sim-multilevel_chains	x86_64		read		0		0x856B008	0x7FFFFFFFFFFFFFFE	N	N	N	ALLOW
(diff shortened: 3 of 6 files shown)


pcmoore:
Hi @andreas-schwab, thanks for the PR.

Given that we now have two arches that we are excluding in test 4 and 6, I'm thinking it might be better to change `open(2)` to `openat(2)` and `stat(2)` to `fstat(2)` as both of those syscalls are supported on all the arches/ABIs that libseccomp currently supports.  Would you mind updating your PR to do this?  If you don't have time, that's okay, just let us know.

* [super handy searchable syscall table](https://github.com/seccomp/libseccomp/blob/master/src/syscalls.csv)

andreas-schwab:
fstat isn't universally available either (it can be called fstatat, fstat64, newfstat), the newest generic default is statx.

pcmoore:
According to the internal libseccomp syscall table `fstat(2)` is available on all arches/ABIs that we currently support, if that isn't correct that is a much bigger issue.  Can you verify that @andreas-schwab?

andreas-schwab:
So what is the correct way to replace open with openat in these tests?

pcmoore:
> So what is the correct way to replace open with openat in these tests?

Run `sed "s/open/openat/"` ;)

Sorry, I couldn't resist a bad joke.  Although in reality I don't think it would be that much more work than that since both test 4 and 6 are "simulated" tests and don't run on live on the system.  Just make sure you replace the `open(2)` calls in both the C, Python, and "*.tests" file.

Here is a *completely* untested example to get you going in the right direction:
```diff
diff --git a/tests/04-sim-multilevel_chains.c b/tests/04-sim-multilevel_chains.c
index a660b40..e3e4f9b 100644
--- a/tests/04-sim-multilevel_chains.c
+++ b/tests/04-sim-multilevel_chains.c
@@ -41,7 +41,7 @@ int main(int argc, char *argv[])
        if (ctx == NULL)
                return ENOMEM;
 
-       rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(open), 0);
+       rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, SCMP_SYS(openat), 0);
        if (rc != 0)
                goto out;
 
diff --git a/tests/04-sim-multilevel_chains.py b/tests/04-sim-multilevel_chains.py
index bcf1ee4..a5127a2 100755
--- a/tests/04-sim-multilevel_chains.py
+++ b/tests/04-sim-multilevel_chains.py
@@ -30,7 +30,7 @@ from seccomp import *
 
 def test(args):
     f = SyscallFilter(KILL)
-    f.add_rule(ALLOW, "open")
+    f.add_rule(ALLOW, "openat")
     f.add_rule(ALLOW, "close")
     f.add_rule(ALLOW, "read",
                Arg(0, EQ, sys.stdin.fileno()),
diff --git a/tests/04-sim-multilevel_chains.tests b/tests/04-sim-multilevel_chains.tests
index 6613f9a..9fdc8a5 100644
--- a/tests/04-sim-multilevel_chains.tests
+++ b/tests/04-sim-multilevel_chains.tests
@@ -8,7 +8,7 @@
 test type: bpf-sim
 
 # Testname                     Arch            Syscall         Arg0            Arg1            Arg2                    Arg3    Arg4    Arg5    Result
-04-sim-multilevel_chains       all,-aarch64    open            0x856B008       4               N                       N       N       N       ALLOW
+04-sim-multilevel_chains       all,-aarch64    openat          0x856B008       4               N                       N       N       N       ALLOW
 04-sim-multilevel_chains       all             close           4               N               N                       N       N       N       ALLOW
 04-sim-multilevel_chains       x86             read            0               0x856B008       0x7FFFFFFE              N       N       N       ALLOW
 04-sim-multilevel_chains       x86_64          read            0               0x856B008       0x7FFFFFFFFFFFFFFE      N       N       N       ALLOW
```

pcmoore:
... oh, and I guess you can remove the "-aarch64" bit from the "*.tests" file since aarch64 supports `openat(2)`.

pcmoore:
... and the "ArgX" values probably need to be shifted as `openat(2)` adds the "dirfd" as the first argument.

andreas-schwab:
It's not that easy, the ranges need to be adjusted as well.

pcmoore:
Like I said, the example was completely untested; it was just intended to get you started.  If the snippet was complete, tested, and working I would have submitted it as a PR myself :)

pcmoore:
Merged into the master branch via a317fabc1fd915f19f7e7326bf7dcb77493f1210 and release-2.5 via cc580a514f05a7fc1f412f66ed002dd8aee89618.  Thanks!

PR #291, opened 2020-08-20
BUG: fix tlive-tsync_notify on 32-bit platforms

karcherm:
the live test tsync_notify expect to be able to push the constant MAGIC as result through the syscall interface. The syscall result is defined as long. The constant 0x1122334455667788UL is forced to be at least unsigned long (even if a smaller type would suffice) by the suffix UL, but will be automatically enlarged to unsigned long long, if unsigned long is not big enough.

Use an explicit cast to remove any bits that do not fit into (unsigned) long.

### DIFF
--- a/tests/58-live-tsync_notify.c
+++ b/tests/58-live-tsync_notify.c
@@ -21,6 +21,7 @@
 
 #include <sys/types.h>
 #include <sys/wait.h>
+#include <asm/unistd.h>
 #include <unistd.h>
 #include <seccomp.h>
 #include <signal.h>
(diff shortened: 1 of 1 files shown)


pcmoore:
Hi @karcherm, I think we fixed this very recently with commit https://github.com/seccomp/libseccomp/commit/bee43d3e884788569860a384e6a38357785a3995, can you test the latest master branch to make sure that works for you?

karcherm:
> 
> 
> Hi @karcherm, I think we fixed this very recently with commit [bee43d3](https://github.com/seccomp/libseccomp/commit/bee43d3e884788569860a384e6a38357785a3995), can you test the latest master branch to make sure that works for you?

That's only for test 51. My pull requests fixes test 58, which has a similar problem. Originally, I patched both 51 and 58, but when porting my patch to current head, i limited it to fixing test 58.

karcherm:
@glaubitz Do you have test results for SH4 at hand that show the need to fix test 58?

pcmoore:
> That's only for test 51. My pull requests fixes test 58, which has a similar problem. Originally, I patched both 51 and 58, but when porting my patch to current head, i limited it to fixing test 58.

Oh, yes, sorry about that!  In an effort to keep things consistent, would you mind modifying your PR so that test 58 is similar to test 51?

karcherm:
> Oh, yes, sorry about that! In an effort to keep things consistent, would you mind modifying your PR so that test 58 is similar to test 51?

Patch adjusted. Testsuite fails on SuperH without this patch, but works with this patch (of course, you need to enable live tests for the testsuite to fail without this patch).

pcmoore:
Merged via f0686d9de911e7ffcdc7364566c1d146e44657c2 and 0336e194e65918ff98b529aa1bab7a01dd8873ea, thanks!

PR #298, opened 2020-10-23
BUG: the seccomp_add_rule(3) and related manpages do not list SCMP_ACT_NOTIFY

drakenclimber:
First cut at documenting SCMP_ACT_NOTIFY in the seccomp_add_rule() manpage.

Here's the human-readable output:

```
       SCMP_ACT_NOTIFY
              A monitoring process will be notified when a process running the seccomp filter calls a syscall  that
              matches the filter rule.  The process that invokes the syscall waits in the kernel until the monitor‐
              ing process has responded via seccomp_notify_respond(3) .

              When a filter utilizing SCMP_ACT_NOTIFY is loaded into the kernel, the kernel generates  a  notifica‐
              tion  fd  that  must  be used to communicate between the monitoring process and the process(es) being
              filtered.  See seccomp_notif_fd(3) for more information.
```

### DIFF
--- a/doc/man/man3/seccomp_rule_add.3
+++ b/doc/man/man3/seccomp_rule_add.3
@@ -209,6 +209,22 @@ matches the filter rule but the syscall will be logged.
 .B SCMP_ACT_ALLOW
 The seccomp filter will have no effect on the thread calling the syscall if it
 matches the filter rule.
+.TP
+.B SCMP_ACT_NOTIFY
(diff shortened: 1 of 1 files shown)


pcmoore:
Looks good to me, merged via 8826022c0b7dd01a09973596750f038334780aad and e201339dc87eac6ce848f441b5b109510ee48dd2.

PR #306, opened 2020-12-14
BUG: use old SECCOMP_IOCTL_NOTIF_ID_VALID number if necessary

sroracle:
[Kernel commit 47e33c05f9f07cac3de833e531bcac9ae052c7ca](https://git.kernel.org/torvalds/c/47e33c05f9f07cac3de833e531bcac9ae052c7ca) changed the
public definition of SECCOMP_IOCTL_NOTIF_ID_VALID for correctness sake
because it had the wrong direction (no current functional change). If
libseccomp is built against kernel headers after this commit but is run
on a kernel that was built prior to this commit, then the ioctl will
always return EINVAL and thus seccomp_notify_id_valid will incorrectly
return -ENOENT.

Copy the (now non-public) definition of the old ioctl number and try it
if the ioctl with the number from the kernel headers fails with -1
EINVAL.

Signed-off-by: Max Rees <maxcrees@me.com>

----

**Note**: should I also update the value of `SECCOMP_IOCTL_NOTIF_ID_VALID` in the `#ifndef SECCOMP_RET_USER_NOTIF` in `src/system.h` while I'm at it?

### DIFF
--- a/src/system.c
+++ b/src/system.c
@@ -535,10 +535,20 @@ int sys_notify_respond(int fd, struct seccomp_notif_resp *resp)
  */
 int sys_notify_id_valid(int fd, uint64_t id)
 {
+	int rc;
 	if (state.sup_user_notif <= 0)
 		return -EOPNOTSUPP;
 
-	if (ioctl(fd, SECCOMP_IOCTL_NOTIF_ID_VALID, &id) < 0)
+	rc = ioctl(fd, SECCOMP_IOCTL_NOTIF_ID_VALID, &id);
+	if (rc < 0 && errno == EINVAL)
+		/* It is possible that libseccomp was built against newer kernel
+		 * headers than the kernel it is running on. If so, the older
+		 * runtime kernel may not support the "fixed"
+		 * SECCOMP_IOCTL_NOTIF_ID_VALID ioctl number which was introduced in
+		 * kernel commit 47e33c05f9f0 ("seccomp: Fix ioctl number for
+		 * SECCOMP_IOCTL_NOTIF_ID_VALID"). Try the old value. */
+		rc = ioctl(fd, SECCOMP_IOCTL_NOTIF_ID_VALID_WRONG_DIR, &id);
+	if (rc < 0)
 		return -ENOENT;
 	return 0;
 }
--- a/src/system.h
+++ b/src/system.h
@@ -179,8 +179,10 @@ struct seccomp_notif_resp {
 #define SECCOMP_IOCTL_NOTIF_RECV        SECCOMP_IOWR(0, struct seccomp_notif)
 #define SECCOMP_IOCTL_NOTIF_SEND        SECCOMP_IOWR(1, \
 						     struct seccomp_notif_resp)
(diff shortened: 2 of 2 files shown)


sroracle:
**Other notes**: the reverse case is not true - compiling against old headers and running against a newer kernel will work fine because the newer kernel still supports the old ioctl number.

This behavior was discovered when running on Linux 5.4.5 after compiling against the 5.4.66 headers. This is because the commit was not backported until 5.4.59 (gregkh/linux@0f09c88f207c0b6a45454e770829a7aba814059e).

pcmoore:
Thanks @sroracle, I've added this to the queue of things for v2.5.2 based on your description but I haven't looked too closely at it yet.  We will get to it, but it's a busy time of year right now :)

src/system.c
```diff
@@ -535,10 +535,18 @@ int sys_notify_respond(int fd, struct seccomp_notif_resp *resp)
...
 	if (state.sup_user_notif <= 0)
 		return -EOPNOTSUPP;
 
-	if (ioctl(fd, SECCOMP_IOCTL_NOTIF_ID_VALID, &id) < 0)
+	rc = ioctl(fd, SECCOMP_IOCTL_NOTIF_ID_VALID, &id);
+	if (rc < 0 && errno == EINVAL)
+		/* We were built against newer kernel headers but are running on an
+		 * older kernel which doesn't accept the fixed ioctl number. Try the
+		 * old one.
+		 * See kernel commit 47e33c05f9f07cac3de833e531bcac9ae052c7ca */
```
pcmoore, line 547: It might be nice to use the kernel standard of a 12-char commit ID followed by the subject line in parentesis and double quotes, e.g. `0123456789ab ("seccomp: let's break compatibility!")`.

drakenclimber:
The change looks correct to me.  As @pcmoore requested, using the standard `47e33c05f9f0 ("seccomp: Fix ioctl number for SECCOMP_IOCTL_NOTIF_ID_VALID")` format in both the code comment and the git commit log would be a good cleanup.  With those small changes, I think it's good to go.

Thanks for the patch, @sroracle.

sroracle:
V2: change comment and commit message wording.

src/system.h
```diff
@@ -181,6 +181,8 @@ struct seccomp_notif_resp {
 						     struct seccomp_notif_resp)
 #define SECCOMP_IOCTL_NOTIF_ID_VALID    SECCOMP_IOR(2, __u64)
```
drakenclimber, line 182: My apologies @sroracle, I forgot about this line.  Yeah, we should probably update `SECCOMP_IOCTL_NOTIF_ID_VALID` to match the kernel.  You're correct.

sroracle:
V3: also update fallback definition of SECCOMP_IOCTL_NOTIF_ID_VALID to the new value.

pcmoore:
Thanks @sroracle, I apologize for the delay.  I've merged this via 83d7b022fa7ef8c24516cc668efc879e5398403f and b926e5af7295b04e8cf7ac4929731fed0f53a711.

PR #308, opened 2021-01-06
RFE: improve test performance

pcmoore:
This PR has two patches intended to speed up the automated regression tests. The first patch adds support for running the tests in parallel, similar to `make -jX`, and the second patch limits the test fuzzing to five iterations for each test, just as we do for the Travis CI runs today.

The individual patch descriptions have more information on the changes and the performance increases for each, but cumulatively this PR drops the test run time from over 14 minutes to just over 3.5 minutes on my eight core laptop.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,7 +69,7 @@ install:
 script:
   - ./configure --enable-python
   - make check-build
-  - LIBSECCOMP_TSTCFG_STRESSCNT=5 make check
+  - make check
   - LIBSECCOMP_TSTCFG_TYPE=live LIBSECCOMP_TSTCFG_MODE_LIST=c make -C tests check
   - |
     if [ $TRAVIS_CPU_ARCH == "amd64" -o -x scan-build ]; then
--- a/tests/01-sim-allow.tests
+++ b/tests/01-sim-allow.tests
@@ -13,7 +13,7 @@ test type: bpf-sim
 test type: bpf-sim-fuzz
 
 # Testname	StressCount
-01-sim-allow	50
+01-sim-allow	5
 
 test type: bpf-valgrind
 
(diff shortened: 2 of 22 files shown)


pcmoore:
Updated this PR with one more patch to make it easier to specify the job count via environment variables (useful for `make check`).  Let me know what you think @drakenclimber, these patches have speed up verification of libseccomp quite significantly on my dev/test system.

drakenclimber:
I did an in-depth review of the changes, and they look good to me.

I do have one small question, though.  On the head of master right now, here's what I see:
```
$ ./regression -m c -m python | wc -l
17873
```

But when I run the same command against this branch, the line count is quite a bit smaller.
```
$ ./regression -m c -m python -j 8 | wc -l
16073
```

I expected the line count to be similar (if not identical.)  Any thoughts?  I don't think it's a rate-limiting issue because I can reproduce these numbers every single time.

pcmoore:
The drop in test count is from commit 4f6b6050641cbd350c83afb2bdb96f168aaffece and is intentional:

> tests: limit the number of bpf-sim-fuzz test iterations
> 
> The bpf-sim-fuzz test type was always intended to be a fuzzer for the generated BPF filters but the current implementation fuzzes the simulator more than the filter itself.  Due to this, let's bump down the bpf-sim-fuzz iterations from 50 to 5, this is what we currently do on Travis CI and it seems to be okay.
> 
> This change drops the number of tests by 1800 and the runtime from 4m7s to 3m35s on my test system running eight parallel jobs.
> 
> Signed-off-by: Paul Moore <paul@paul-moore.com>

:)

drakenclimber:
> The drop in test count is from commit [4f6b605](https://github.com/seccomp/libseccomp/commit/4f6b6050641cbd350c83afb2bdb96f168aaffece) and is intentional:

Thanks.  Duh!  I spent so much time focusing on the log management and parallelization, I missed the completely obvious.

This is a really good improvement.  Thanks for tackling it.

drakenclimber (approved):

PR #317, opened 2021-02-16
BUG: doc/man: don't return -1 from main

kolyshkin:
It is custom for C functions to return a negative integer to denote
an error. This should not be done with main() though, as the return
value is used as an exit code, which is a byte (unsigned char), so
return -1 becomes 255, which is unconventional.

Fix examples where main returns -1 to return 1.


drakenclimber (approved):
Looks good to me.  Thanks @kolyshkin 

rhatdan:
LGTM

pcmoore:
Sorry for the delay!  I've merged this via 6d5a0bb22c8162c167900a47d57ebdeeb22ff8c0 and f679355a87d8e8ad5bc16d23e2a2e174a03a6f11 - thank you!

Issue #12, opened 2015-08-24
RFE: add "last known good" test BPF output to the tree for regression testing


pcmoore:
We have some basic tooling to generate (tests/testgen) and compare (tests/testdiff) BPF output from multiple test runs, we should investigate adding known good BPF output from the tests to the tree to help catch problems with BPF generation.

These "known good" outputs will need to be maintained/updated as improvements are made to libseccomp, but the overhead shouldn't be prohibitive.

Issue #11, opened 2015-08-21
RFE: Support "maximum kernel version"


cgwalters:
As system calls are added to the kernel, I feel there is not enough discussion by default of the wide variety of applications that will suddenly gain access to a new attack surface.

The canonical example here is `perf_event_open()`, the source of numerous CVEs.  While perf is awesome, my (e.g.) web server should not (by default) be able to use it.

It's possible to use seccomp today to blacklist.  whitelists can get very difficult to manage.

One thing that might be useful is a filter for any system calls newer than a particular kernel version, say 3.10.  That way, each new system call would have to be verified for use in e.g. containers before it's added.  Upgrading the kernel wouldn't suddenly expose containers to new attack surface.

In a discussion with @pcmoore he indicated this could be another annotation in the struct in e.g. `arch-x86-syscalls.c`.

nmav:
+1
That would help make blacklists usable for mitigation of security issues.

pcmoore:
@nmav to be clear, this RFE is for adding information to the internal syscall tables about when the syscall was first introduced to the Linux kernel, not for adding logic to determine if the current running kernel supports a given syscall.  However, if you are trying to block a syscall, you can do so with libseccomp regardless of if it is supported on a particular arch/ABI and kernel version, libseccomp will do the right thing for you.

pcmoore:
This RFE is almost five years old, and outside of a single discussion with @cgwalters I haven't seen or heard of much other interest in such a feature.  With plenty of other open issues, most with higher priority, it is not clear when we would work on this, or even if such a thing would be a useful addition.

@cgwalters and @drakenclimber what do you think of this issue in 2020?  I'm tempted to close this as WONTFIX, but I would like to get some comments and feedback before we take that step.

drakenclimber:
> @cgwalters and @drakenclimber what do you think of this issue in 2020? I'm tempted to close this as WONTFIX, but I would like to get some comments and feedback before we take that step.

Honestly I think this is a really cool idea.  Several of my in-house customers are using allowlists because of this exact reason.  If they were to use a denylist and a new syscall is added to the kernel, then that syscall would be another avenue of attack.

Let's leave it open for a bit longer.  I'll ask around within Oracle and see if any customers are interested enough in this feature for me to pick it up.  But @cgwalters (or anyone else for that matter) is totally welcome to own it if they have the time and interest :).

pcmoore:
[...]

PR #319, opened 2021-03-18
BUG: fix uninitialized value usage in gen_bpf.c

giuseppe:
it was reported by clang with the option -fsanitize=memory:

Uninitialized bytes in MemcmpInterceptorCommon at offset 0 inside [0x7070000002a0, 56)
==3791089==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x482a2c in memcmp (fuzzer+0x482a2c)
    #1 0x7fed2f120ebb in _hsh_add src/libseccomp/src/gen_bpf.c:598:9
    #2 0x7fed2f121715 in _gen_bpf_action_hsh src/libseccomp/src/gen_bpf.c:796:6
    #3 0x7fed2f121a53 in _gen_bpf_node src/libseccomp/src/gen_bpf.c:831:11
    #4 0x7fed2f121a53 in _gen_bpf_chain.isra.0 src/libseccomp/src/gen_bpf.c:1072:13
    #5 0x7fed2f121f16 in _gen_bpf_chain_lvl_res src/libseccomp/src/gen_bpf.c:977:12
    #6 0x7fed2f121c74 in _gen_bpf_chain.isra.0 src/libseccomp/src/gen_bpf.c:1124:12
    #7 0x7fed2f12253c in _gen_bpf_syscall src/libseccomp/src/gen_bpf.c:1520:10
    #8 0x7fed2f12253c in _gen_bpf_syscalls src/libseccomp/src/gen_bpf.c:1615:18
    #9 0x7fed2f12253c in _gen_bpf_arch src/libseccomp/src/gen_bpf.c:1683:7
    #10 0x7fed2f12253c in _gen_bpf_build_bpf src/libseccomp/src/gen_bpf.c:2056:11
    #11 0x7fed2f12253c in gen_bpf_generate src/libseccomp/src/gen_bpf.c:2321:7
    #12 0x7fed2f11f41c in seccomp_export_bpf src/libseccomp/src/api.c:724:7

  Uninitialized value was created by a heap allocation
    #0 0x4547ef in realloc (fuzzer+0x4547ef)
    #1 0x7fed2f121244 in _blk_resize src/libseccomp/src/gen_bpf.c:362:8
    #2 0x7fed2f121244 in _blk_append src/libseccomp/src/gen_bpf.c:394:6

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

### DIFF
--- a/src/gen_bpf.c
+++ b/src/gen_bpf.c
@@ -351,6 +351,7 @@ static struct bpf_blk *_blk_resize(struct bpf_state *state,
 {
 	unsigned int size_adj = (AINC_BLK > size_add ? AINC_BLK : size_add);
 	struct bpf_instr *new;
+	size_t old_size, new_size;
 
 	if (blk == NULL)
 		return NULL;
@@ -358,8 +359,10 @@ static struct bpf_blk *_blk_resize(struct bpf_state *state,
 	if ((blk->blk_cnt + size_adj) <= blk->blk_alloc)
 		return blk;
 
+	old_size = blk->blk_alloc * sizeof(*new);
 	blk->blk_alloc += size_adj;
-	new = realloc(blk->blks, blk->blk_alloc * sizeof(*(blk->blks)));
+	new_size = blk->blk_alloc * sizeof(*new);
+	new = zrealloc(blk->blks, old_size, new_size);
 	if (new == NULL) {
 		_blk_free(state, blk);
 		return NULL;
@@ -443,10 +446,13 @@ static int _bpf_append_blk(struct bpf_program *prg, const struct bpf_blk *blk)
 	bpf_instr_raw *i_iter;
 	unsigned int old_cnt = prg->blk_cnt;
 	unsigned int iter;
+	size_t old_size, new_size;
 
 	/* (re)allocate the program memory */
+	old_size = BPF_PGM_SIZE(prg);
 	prg->blk_cnt += blk->blk_cnt;
-	i_new = realloc(prg->blks, BPF_PGM_SIZE(prg));
+	new_size = BPF_PGM_SIZE(prg);
+	i_new = zrealloc(prg->blks, old_size, new_size);
 	if (i_new == NULL) {
 		rc = -ENOMEM;
 		goto bpf_append_blk_failure;
--- a/src/helper.c
+++ b/src/helper.c
@@ -34,18 +34,11 @@
  */
 void *zmalloc(size_t size)
 {
-	void *ptr;
-
 	/* NOTE: unlike malloc() zero size allocations always return NULL */
 	if (size == 0)
 		return NULL;
 
-	ptr = malloc(size);
-	if (!ptr)
-		return NULL;
-	memset(ptr, 0, size);
-
-	return ptr;
+	return calloc(1, size);
 }
 
 /**
--- a/src/helper.c
+++ b/src/helper.c
@@ -47,3 +47,28 @@ void *zmalloc(size_t size)
 
 	return ptr;
 }
+
+/**
+ * Change the size of an allocated buffer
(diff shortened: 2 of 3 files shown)


pcmoore:
I could be wrong, but I suspect there is a gap somewhere in `struct bpf_blk` that isn't normally written to but it is read when `_hsh_add_src()` runs over the entire struct, triggering the warning/error.

@giuseppe there is already a libseccomp function, `zmalloc()` in "src/helper.c" which does a `malloc() + memset()`; how about refactoring your approach in this PR to create a `zrealloc()` that does a similar thing and the converting the handful of existing calls to `realloc()`?  Of course the `zrealloc()` implementation would need to take an additional argument to represent the size of the current buffer, but that shouldn't be too hard to sort out.

giuseppe:
> @giuseppe there is already a libseccomp function, `zmalloc()` in "src/helper.c" which does a `malloc() + memset()`; how about refactoring your approach in this PR to create a `zrealloc()` that does a similar thing and then converting the handful of existing calls to `realloc()`? Of course the `zrealloc()` implementation would need to take an additional argument to represent the size of the current buffer, but that shouldn't be too hard to sort out.

sure, I can do that.

Are you fine if I refactor `zmalloc` to use `calloc` instead of `malloc` (or possibly just  replace `zmalloc` with `calloc`)?

AFAIK, `calloc` avoids the additional `memset` when the std library knows the memory is already set to 0.

giuseppe:
pushed a new version that adds `zrealloc`.

pcmoore:
This looks much better, thank you!  Would you also mind converting the other `realloc()` callers?
```
% grep "realloc" $(find ./src -name "*.c")
./src/gen_bpf.c:        new = realloc(blk->blks, blk->blk_alloc * sizeof(*(blk->blks)));
./src/gen_bpf.c:        i_new = realloc(prg->blks, BPF_PGM_SIZE(prg));
./src/db.c:     dbs = realloc(col_dst->filters,
./src/db.c:     dbs = realloc(col->filters,
./src/db.c:             /* NOTE: if we can't do the realloc it isn't fatal, we just
./src/db.c:             dbs = realloc(col->filters,
./src/db.c:             tmp_f = realloc(snap->filters,
```

> Are you fine if I refactor zmalloc to use calloc instead of malloc (or possibly just replace zmalloc with calloc)?

Sure, that sounds good.  You can do it in this PR if you like, but please do it in a separate patch from the realloc work.

giuseppe:
> ./src/db.c:     dbs = realloc(col_dst->filters,
> ./src/db.c:     dbs = realloc(col->filters,
> ./src/db.c:             /* NOTE: if we can't do the realloc it isn't fatal, we just
> ./src/db.c:             dbs = realloc(col->filters,
> ./src/db.c:             tmp_f = realloc(snap->filters,

I've not touched these calls because the new allocated space is immediately written after the `realloc`, so there won't be any advantage with `zrealloc`, and it would just be more expensive.

If you still want it, I can change these occurrences as well.

pcmoore:
> I've not touched these calls because the new allocated space is immediately written after the realloc, so there won't be any advantage with zrealloc, and it would just be more expensive.

Fair enough.  There could be an argument about doing it anyway for a belt-and-suspenders approach, but looking at the code that is a pretty weak argument in those cases.  Let's leave it as-is.

`Acked-by: Paul Moore <paul@paul-moore.com>`

pcmoore (approved):

drakenclimber:
Looks good to me.  Thanks, @giuseppe 

giuseppe:
is this PR fine to merge?

drakenclimber (approved):

drakenclimber:
Sorry about that, @giuseppe.  I forgot to click the "approve" button.  I'll merge it in a second

PR #322, opened 2021-04-17
BUG: update the syscall table to v5.12-rc7

pcmoore:
Due to additional ABIs in main we can't do a simple backport or copy
of the syscall table so we are generating it directly in the
release-2.5 branch.

This patch also fixes the missing faccessat2() #defines in the
seccomp-syscalls.h header file.


pcmoore:
@drakenclimber given you ok'd the same fixes for the main branch I'm going to assume this also looks good to you, if not let me know and we can revert/fix.

Merged at c56a00fe173a7dd5a8326431ae28863ce432bbc1.

Issue #299, opened 2020-11-06
BUG: look into replacing Travis CI with GitHub actions


pcmoore:
There are plenty of articles on this, but see The Register article below for some background on this issue:

* https://www.theregister.com/2020/11/02/travis_ci_pricng

With Travis CI becoming potentially unreliable for libseccomp, we should investigate moving out CI testing to GitHub actions:

* https://docs.github.com/en/free-pro-team@latest/actions

drakenclimber:
Here's a guide for migrating from TravisCI to Github Actions.  I am hoping to experiment with this over the next few days
https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/migrating-from-travis-ci-to-github-actions

drakenclimber:
I'm really liking the look of Github Actions.  I just sent out a patchset to switch libcgroup from Travis CI to Github Actions.

https://github.com/drakenclimber/libcgroup/tree/issues/github_actions
https://sourceforge.net/p/libcg/mailman/message/37165461/

I'll try to put together a patchset to transition libseccomp this week.

pcmoore:
That sounds good @drakenclimber, thanks!

drakenclimber:
Here's my current status.  I have Github Actions [working on amd64](https://github.com/drakenclimber/libseccomp/actions/runs/396925419) with some slight deviations from our current Travis CI solution, but I have concerns about other architectures.

Pros:
- Easy to use.  I was able to quickly setup a [continuous integration workflow](https://github.com/drakenclimber/libseccomp/blob/wip/github_actions2/.github/workflows/continuous-integration.yml) in Github Actions
- It's also easy to encapsulate operations into an [action](https://github.com/drakenclimber/libcgroup/blob/issues/github_actions/.github/actions/setup-libcgroup/action.yml).  These actions can then be reused within the workflow.  (Strangely an [action cannot invoke another action](https://github.community/t/call-an-action-from-another-action/17524).)
- Acceptable speed (for amd64 at least).  The entire test suite completed in [22 minutes](https://github.com/drakenclimber/libseccomp/runs/1488740223)
- [Good GUI](https://github.com/drakenclimber/libseccomp/runs/1488740223) and failing steps are clearly delineated
[...]

Issue #312, opened 2021-01-19
RFE: Transition from travis-ci.org to travis-ci.com


drakenclimber:
travis-ci.org is being decommissioned and will go away in the near future.

Follow the steps here to transition libseccomp to the new travis-ci.com site.
https://docs.travis-ci.com/user/migrate/open-source-repository-migration

drakenclimber:
I logged into travis-ci.com, but they are still in the beta phase, thus we can't transition yet.

pcmoore:
Bah!  They are definitely not making this easy are they? ;)

pcmoore:
Since the transition deadline is still "fuzzy" to put it mildly I'm going to pull this from the v2.5.2 milestone and let it float.  We will still need to look at this at some point, but until Travis makes the move there is nothing for us to do here.

pcmoore:
I just went to verify that the Travis build completed okay after some pushes and I was greeted by this message:
> Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on.

... it looks like we need to figure this out, and soon.

drakenclimber:
I'm a little leery of Travis' new business model.  It looks like open source will remain free (for now), but they have not made it easy.  We may need to periodically request more "credits" from them.
https://docs.travis-ci.com/user/billing-faq/#what-if-i-am-building-open-source

```
What if I am building open source? #

Each of the Travis CI Plans contains an amount of special OSS credits per month assigned to
run builds only on public repositories. To find out more about it please contact the Travis CI
support team. In the email please include:

* Your account name and your VCS provider (like travis-ci.com/github/[your account name] )
* How many credits (build minutes) you’d like to request (should your run out of credits again
  you can repeat the process to request more or to discuss a renewable amount)
```

pcmoore:
Ugh, that sucks :/

I need to refresh my memory on what you found out about GH Actions; it was far from perfect, but considering the Travis CI changes it might be the better option.

pcmoore:
Here we go ... #299

pcmoore:
Okay, having refreshed my memory on #299 I think the best option *right now* is to migrate to GH Actions and stick just to x86_64 CI builds for the time being.  That seems like the quickest path to restoring PR/branch and code coverage tests with the least amount of headaches.

[...]

PR #329, opened 2021-07-21
RFE: Transition from TravisCI to Github Actions

drakenclimber:
This pull request resolves Issue #299 and Issue #312.

This patchset transitions the libseccomp continuous integration from TravisCI to Github Actions.  Here's a brief rundown of the pros/cons of each CI system:

| | TravisCI | Github Actions |
| -------------- | -------------- | -------------- |
| Run Time<sup>1</sup> | ~24 min | ~17 min |
| Architectures Being Tested<sup>2</sup> | amd64, arm64, ppc64le | amd64 |
| Parallelization | Each architecture is a separate run | Each test type (simulation, live, scan build, code coverage) is a separate run |
| Reliability | Excellent | Excellent |
| When Does it Run | On commits and pull requests | On commits and pull requests |
| Costs | Uncertain.  Travis is still claiming [free support](https://docs.travis-ci.com/user/billing-faq/#what-if-i-am-building-open-source) for open source, but it looks both tedious and tenuous | Free.  We've had no issues in libcgroup so far |
| User Experience - debugging failures | Must wade through the logs.  Likely tedious for occasional libseccomp contributors | Each test type is split into a separate run, and the [GUI clearly delineates them](https://github.com/drakenclimber/libseccomp/runs/3125400036).  A live test failure vs. a simulation failure is obvious, but a user still needs to dig into logs for the exact cause |
| Code Coverage<sup>3</sup> | Our current TravisCI code coverage uses a [user-generated coveralls wrapper](https://github.com/eddyxu/cpp-coveralls/blob/master/README.md) that sends the coverage data to coveralls.io.  It's worked well, but is a little nonstandard | The proposed Github Actions solution uses a [Github Action developed by Coveralls](https://github.com/coverallsapp/github-action), and in theory has better support and integration with Github Actions and Coveralls. |
| Current Code Coverage Numbers | [93.0%](https://coveralls.io/github/seccomp/libseccomp?branch=main) | [89.9%](https://coveralls.io/builds/41545661) |

<sup>1</sup> It seems that the TravisCI VMs are more powerful than the Github Actions VMs, but Github Actions makes parallelization much easier.  Thus the nontrivial run time differences.
<sup>2</sup> There are a few ways to add support for additional architectures to Github Actions via either [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow) or possibly this [Docker container solution](https://github.com/marketplace/actions/run-on-architecture)
<sup>3</sup> The code coverage numbers differ slightly from the TravisCI implementation to the Github Actions implementation.  I think the Github Actions numbers are more accurate, but I haven't dug too deeply.  (Also, I can't get Github Actions code coverage to ignore `arch-syscall-check.c`.  It's kinda bugging me :).)


pcmoore:
I haven't looked at the changes/commits yet, but +1000 on the ultra snazzy PR description!

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,40 @@
...
+
+name: Setup the libseccomp directory
+description: "Install dependencies and configure libseccomp"
+runs:
+  using: "composite"
+  steps:
+  - run: sudo apt-get install build-essential valgrind clang-tools lcov gperf astyle codespell
+    shell: bash
+  - run: |
+      sudo apt-get install -y python3.6 python3-distutils python3-setuptools python3-pip
```
pcmoore, line 30: Can we specify the latest Python v3 instead of specifically Python v3.6?  For example, does it work if you attempt to install the `python3` package?

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,40 @@
...
+      python3 -m pip install --upgrade pip
+      python3 -m pip install cython
+      # Add cython to the path
+      echo "/home/runner/.local/bin" >> $GITHUB_PATH
```
pcmoore, line 34: Does `echo "$HOME/.local/bin" >> $GITHUB_PATH` also work?  It would be nice not to have to rely on a hardcoded username.

pcmoore (changes requested):

pcmoore:
@drakenclimber out of curiosity, what is the code coverage percentage with the GH Actions approach?

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,92 @@
...
+    runs-on: ubuntu-20.04
+
+    steps:
+    - name: Checkout from github
+      uses: actions/checkout@v2
+    - name: Initialize libseccomp
+      uses: ./.github/actions/setup
+    - name: Run scan-build
+      run: |
+        make clean
```
pcmoore, line 61: It's probably best to run `configure` before `make clean`.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,92 @@
...
+    runs-on: ubuntu-20.04
+
+    steps:
+    - name: Checkout from github
+      uses: actions/checkout@v2
+    - name: Initialize libseccomp
+      uses: ./.github/actions/setup
+    - name: Run tests with code coverage enabled
+      run: |
+        make clean
```
pcmoore, line 76: Same as above, do the `configure` before `make clean`.

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,40 @@
...
+    shell: bash
+  - run: |
+      ./autogen.sh
+      ./configure --enable-python
+      make check-build
```
pcmoore, line 39: I don't think there is any doubt that running `./autogen.sh` is the right thing to do here, but I'm less convinced about the `configure` and `make` commands; should we relocate those to the individual CI jobs, even if it results in some duplication among separate jobs?

drakenclimber: I completely agree.  For the initial version, I tried to keep the Github Actions logic the same as the Travis logic.

But since the overall design is so much different (runs are split up by test type), your suggestion is definitely a better solution.

.github/actions/setup/action.yml
drakenclimber: Good call.  Looks like ubuntu supports `python3` like you mentioned.  Thanks.

drakenclimber:
> @drakenclimber out of curiosity, what is the code coverage percentage with the GH Actions approach?

I added a link to each in the table above.
[93.0%](https://coveralls.io/github/seccomp/libseccomp?branch=main) - TravisCI
[89.9%](https://coveralls.io/builds/41545661) - Github Actions

pcmoore:
> > @drakenclimber out of curiosity, what is the code coverage percentage with the GH Actions approach?
> 
> I added a link to each in the table above.
> [93.0%](https://coveralls.io/github/seccomp/libseccomp?branch=main) - TravisCI
> [89.9%](https://coveralls.io/builds/41545661) - Github Actions

Arrrgh less than 90%?!  Several years ago I spent an absurd amount of time one year getting the coverage above 90%, seeing it drop below 90% is a sad moment :)  Although most of that was due to fixing various small bugs that cropped up during the additional testing so I can't be too upset.

Although looking at the coverage data there are definitely a few things that look .... off.  For example, the file list is definitely not correct (files missing, others added that shouldn't be there, etc.).  Regardless, that's all something we can fix later.

pcmoore:
Merged at HEAD 04245d933fad94625b45a4d100112a3483ed9292.

I'm thinking that maybe we keep this issue open until we sort out some of the remaining code coverage issues and backport everything to the release-2.5 branch, or should we close this out and create a new issue for that - thoughts?

pcmoore:
With this merged into the main branch and PR #331 backporting this to the release-2.5 branch I think we can close this out now, any remaining issues should be treated as new issues or handled in the backport PR.

PR #339, opened 2021-09-24
doc: seccomp_rule_add.3: add -EACCES return value

kolyshkin:
The `-EACCES` return value from `seccomp_rule_add` and friends was added
by commit 83989be02 (included into 2.5.0), which tells that this is "part of our
... API promise", so it needs to be documented accordingly. Add it.

_The discussion leading to this PR is in https://github.com/seccomp/libseccomp-golang/pull/74,
however we did not came to the agreement whether `-EACCES` is the best choice._

Fixes: 83989be02


pcmoore:
Thanks @kolyshkin.

I'm still wondering if we are better off returning -EACCES in this case or changing it to -EEXIST, but I guess sticking with -EACCES is probably for the best as that is what we have done starting with v2.5.x and the beginning of our "API promise".

My next question is if we want to be a specific as this patch when explaining the reason for the error code, *"The action argument equals the default action of the filter"* is pretty specific.  Perhaps consider changing it to: *"The rule conflicts with the default filter settings"*?

Thoughts?

kolyshkin:
> My next question is if we want to be a specific as this patch when explaining the reason for the error code, _"The action argument equals the default action of the filter"_ is pretty specific. Perhaps consider changing it to: _"The rule conflicts with the default filter settings"_?
> 
> Thoughts?

What about something like _"The rule conflicts with the filter (for example, the rule action equals the default action of the filter)"_? This way we are specific (and clear) but still leave room for other similar errors.

drakenclimber:
> > My next question is if we want to be a specific as this patch when explaining the reason for the error code, _"The action argument equals the default action of the filter"_ is pretty specific. Perhaps consider changing it to: _"The rule conflicts with the default filter settings"_?
> > Thoughts?
> 
> What about something like _"The rule conflicts with the filter (for example, the rule action equals the default action of the filter)"_? This way we are specific (and clear) but still leave room for other similar errors.

I'm fine with either of these.

kolyshkin:
Updated to
```
       -EACCCES
              The rule conflicts with the filter (for example, the rule action
              equals the default action of the filter).
```

and fix a typesetting error I made earlier (missing `.TP`).

PR #344, opened 2021-10-15
RFE: check the -s flag value in regression script

Cypresslin:
Add help message to explain how to assign multiple tests with the -s
flag.

Also, check the input value for the -s flag to make sure it's in a
proper format. Example:

    $ ./regression -s 11-basic-basic_errors
    Please just enter number for the -s flag, terminating...

Signed-off-by: Po-Hsu Lin \<po-hsu.lin@canonical.com\>


pcmoore:
Please see my comment in https://github.com/seccomp/libseccomp/issues/345, I think you are confusing the '-b' and '-s' options.

Cypresslin:
Yes I got confused  by the definition of "batch of tests" and "individual test number".
This can be closed.
Thanks!

PR #351, opened 2021-10-27
BUG: sync the src/syscalls.csv and include/seccomp-syscalls.h files

pcmoore:
Unfortunately the internal syscall table and the exported syscall header file have become disconnected which can lead to problems where a syscall is not accessible to libseccomp despite being present in the syscall table.  This PR not only corrects the two files, but updates the existing "src/arch-syscall-check" build time test to verify that the two files are in sync.


pcmoore:
> I admit I'm a bash hack ...

I'm *very much* a bash hack :)

Merged into main at f046fd21e3274541021bff1f869bf2c9ef8d0b86 and release-2.5 at fa85b21a592c82b33eded2ec9758329afe483648.

Issue #372, opened 2022-03-16
BUG: Python distutils is deprecated


drakenclimber:
On Fedora 35, the following warning is generated when building the Python bindings:
```
./setup.py:26: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
```

pcmoore:
The libcgroup approach looks good to me.  I assume this works on older versions of Python/Cython as well?

drakenclimber:
> The libcgroup approach looks good to me. I assume this works on older versions of Python/Cython as well?

I haven't extensively tested the limits of which python/cython versions are supported, but I do know that it works on Oracle Linux 7 which contains:
```
$ python --version
Python 3.6.8
$ cython --version
Cython version 0.29.28
```

PR #374, opened 2022-03-21
BUG: Python distutils is deprecated

drakenclimber:
The python distutils package is deprecated.  Utilize
setuptools and cythonize instead.

	./setup.py:26: DeprecationWarning: The distutils
	package is deprecated and slated for removal in
	Python 3.12. Use setuptools or check PEP 632 [1] for
	potential alternatives

[1] https://peps.python.org/pep-0632/

Fixes: https://github.com/seccomp/libseccomp/issues/372

### DIFF
--- a/src/python/setup.py
+++ b/src/python/setup.py
@@ -23,9 +23,9 @@
 
 import os
 
-from distutils.core import setup
-from distutils.extension import Extension
-from Cython.Distutils import build_ext
+from setuptools import setup
+from setuptools.extension import Extension
+from Cython.Build import cythonize
 
 setup(
 	name = "seccomp",
@@ -37,12 +37,9 @@ setup(
 	maintainer_email = "paul@paul-moore.com",
 	license = "LGPLv2.1",
 	platforms = "Linux",
-	cmdclass = {'build_ext': build_ext},
-	ext_modules = [
+	ext_modules = cythonize([
 		Extension("seccomp", ["seccomp.pyx"],
 			# unable to handle libtool libraries directly
-			extra_objects=["../.libs/libseccomp.a"],
-			# fix build warnings, see PEP 3123
-			extra_compile_args=["-fno-strict-aliasing"])
-	]
+			extra_objects=["../.libs/libseccomp.a"]),
+	])
 )


src/python/setup.py
```diff
@@ -37,12 +37,9 @@
 	maintainer_email = "paul@paul-moore.com",
 	license = "LGPLv2.1",
 	platforms = "Linux",
-	cmdclass = {'build_ext': build_ext},
-	ext_modules = [
+	ext_modules = cythonize([
 		Extension("seccomp", ["seccomp.pyx"],
 			# unable to handle libtool libraries directly
-			extra_objects=["../.libs/libseccomp.a"],
-			# fix build warnings, see PEP 3123
```
drakenclimber, line 45: I didn't observe any build warnings on Python 3.10.2.  Do we still need this?

pcmoore: I ... don't know?  The python bindings went in so long ago I don't recall why it was necessary beyond the comment.

Is it so that the bindings are built with the library statically linked as opposed to using a shared library installed on the system?

jhenstridge: You shouldn't need `-fno-strict-aliasing` with Python 3. It was an issue with Python 2's C API, where the first fields in an extension type's struct duplicated the fields from PyObject (i.e. the reference count and type pointer). In Python 3, the first field of an extension type's struct is a PyObject.

From the look of it, you've dropped Python 2 support, so you could safely drop `-fno-strict-aliasing` too.

pcmoore: Thanks @jhenstridge, mystery solved :)

pcmoore (approved):
Looks good to me.
```
```

pcmoore:
I'm hoping @drakenclimber wasn't sitting on this for a reason, but I'm in a daring mood right now so I'm going to go ahead and merge it ;)  @drakenclimber if you're not ready for this to go in let me know and we can revert it.

Merged via afbde6ddaec7c58c3b281d43b0b287269ffca9bd.

drakenclimber:
I'm good with merging it.   Thanks, @pcmoore.

Got sidetracked with other work, then Covid got me.  I'm getting back up to speed though :)

PR #377, opened 2022-04-05
RFE: update the syscall list for Linux v5.17

pcmoore:
Subject says it all ...


pcmoore:
Merged as this was a trivial syscall update.

PR #395, opened 2022-11-24
RFE: add vgetrandom_alloc()

zx2c4:
This is required to use vDSO-accelerated getrandom().

Don't merge this yet. It hasn't landed upstream, and as of writing is at v17: https://lore.kernel.org/lkml/20240614190646.2081057-1-Jason@zx2c4.com/


drakenclimber:
Per @zx2c4's comment, I'm moving this to draft.  @zx2c4 - feel free to move this back to "ready to review" when it lands upstream.  Thanks

pcmoore:
@drakenclimber @zx2c4 given the (relative) ease of updating the libseccomp syscall tables these days, one might as well do a full syscall table update instead of adding just a single syscall.

For example, the following can be used to update the syscall table for the kernel sources checked out in `<kernel_source_dir>`.  The only gotcha is that syscalls that are not universally supported on all arches need a manual `PNR` define in "include/seccomp-syscalls.h".

```
% make check-build
% cd src
% ./arch-syscall-validate -c syscall.csv <kernel_source_dir>
```

pcmoore:
@zx2c4 any updates on this?  I'm tempted to close this out as it looks like it still hasn't landed in Linus' tree and we would normally just pick this up via the usual syscall update process, but it would be nice to hear from you first.

zx2c4:
No not yet. I'll let you know when it is.

zx2c4:
The syscall side of things wasn't required in the end.

pcmoore:
Thanks for this regardless.  It ended up just being a new `mmap()` flag, yes?

zx2c4:
https://lwn.net/SubscriberLink/983186/d37da11dfd53f1c4/

PR #402, opened 2023-02-24
BUG: Couple of small fixes to LoongArch64 support

xen0n:
No functional change intended, I've found out them while working at one of my pet projects tonight.

also cc @yetist

### DIFF
--- a/src/arch.c
+++ b/src/arch.c
@@ -64,7 +64,7 @@ const struct arch_def *arch_def_native = &arch_def_x86_64;
 const struct arch_def *arch_def_native = &arch_def_arm;
 #elif __aarch64__
 const struct arch_def *arch_def_native = &arch_def_aarch64;
-#elif __loongarch64
+#elif __loongarch_lp64
 const struct arch_def *arch_def_native = &arch_def_loongarch64;
 #elif __mips__ && _MIPS_SIM == _MIPS_SIM_ABI32
 #if __MIPSEB__
--- a/tools/util.c
+++ b/tools/util.c
@@ -44,7 +44,7 @@
 #define ARCH_NATIVE		AUDIT_ARCH_ARM
 #elif __aarch64__
 #define ARCH_NATIVE		AUDIT_ARCH_AARCH64
-#elif __loongarch64
+#elif __loongarch_lp64
 #define ARCH_NATIVE		AUDIT_ARCH_LOONGARCH64
 #elif __mips__ && _MIPS_SIM == _MIPS_SIM_ABI32
 #if __MIPSEB__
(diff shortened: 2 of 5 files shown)


src/arch.c
```diff
@@ -64,7 +64,7 @@ const struct arch_def *arch_def_native = &arch_def_x86_64;
 const struct arch_def *arch_def_native = &arch_def_arm;
 #elif __aarch64__
 const struct arch_def *arch_def_native = &arch_def_aarch64;
-#elif __loongarch64
+#elif __loongarch__ && __loongarch_grlen == 64
```
xry111, line 67: `__loongarch_lp64`, IMO.  We are caring about the syscall ABI which is an ABI anyway, not the length of the HW register.

drakenclimber: I briefly looked through the linked reference, and I think @xry111's comment makes sense.  But I will defer to loongarch experts if there's a better answer

yetist: I think `__loongarch_lp64` should be used here, please refer to this [link](https://loongson.github.io/LoongArch-Documentation/LoongArch-toolchain-conventions-EN.html#_cc_preprocessor_built_in_macro_definitions), Table 14 and Table 15.

tools/util.c
```diff
@@ -44,7 +44,7 @@
 #define ARCH_NATIVE		AUDIT_ARCH_ARM
 #elif __aarch64__
 #define ARCH_NATIVE		AUDIT_ARCH_AARCH64
-#elif __loongarch64
+#elif __loongarch__ && __loongarch_grlen == 64
```
yetist, line 47: `__loongarch_lp64` here.

src/arch.c
xen0n: Okay. I chose to migrate to the `__loongarch_grlen` approach because it's trivial to verify against [the *Toolchain Conventions*](https://loongson.github.io/LoongArch-Documentation/LoongArch-toolchain-conventions-EN.html) that use the same wording. Of course thinking about the actual intent here `__loongarch_lp64` is more appropriate because what we actually care is the ABI; I'll update the branch shortly. Thanks for the reviews!

xen0n:
Sorry for the delay; I've now updated the branch.

drakenclimber (approved):
Looks good to me.  Thanks, @xen0n.

xen0n:
> My apologies everyone, other tasks have kept me very busy lately, but I plan to spend more time on libseccomp later today/tomorrow.  Thanks for your patience!

No worries. I'm also multitasking between the day job and multiple open-source projects (linux, llvm, go and rust this week) so I probably can feel ;-)

xen0n:
I've rebased the code and fixed the CI failure (misspellings of "pseudo") by the way.

pcmoore (approved):

pcmoore:
Merged at 5aceb9d68a14e055c7a702483d612b537244e276, thanks everyone!

PR #419, opened 2023-11-04
BUG: fix detection of WAIT_KILLABLE_RECV flag

alip:
The kernel returns EINVAL when this flag is passed to seccomp without the new listener flag so we should pass this flag along as well.

Reference on where the check happens in the kernel:
https://github.com/torvalds/linux/blob/aea6bf908d730b01bd264a8821159db9463c111c/kernel/seccomp.c#L1926-L1932

### DIFF
--- a/src/system.c
+++ b/src/system.c
@@ -311,8 +311,16 @@ int sys_chk_seccomp_flag(int flag)
 			state.sup_flag_tsync_esrch = _sys_chk_flag_kernel(flag);
 		return state.sup_flag_tsync_esrch;
 	case SECCOMP_FILTER_FLAG_WAIT_KILLABLE_RECV:
-		if (state.sup_flag_wait_kill < 0)
-			state.sup_flag_wait_kill = _sys_chk_flag_kernel(flag|SECCOMP_FILTER_FLAG_NEW_LISTENER);
+		if (state.sup_flag_wait_kill < 0) {
+			/* kernel requires NEW_LISTENER with WAIT_KILLABLE_RECV */
+			flag |= SECCOMP_FILTER_FLAG_NEW_LISTENER;
+			sys_chk_seccomp_flag(SECCOMP_FILTER_FLAG_NEW_LISTENER);
+			if (state.sup_flag_new_listener) {
+				state.sup_flag_wait_kill = _sys_chk_flag_kernel(flag);
+			} else {
+				state.sup_flag_wait_kill = 0;
+			}
+		}
 		return state.sup_flag_wait_kill;
 	}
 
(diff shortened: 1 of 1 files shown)


src/system.c
```diff
@@ -312,7 +312,7 @@ int sys_chk_seccomp_flag(int flag)
 		return state.sup_flag_tsync_esrch;
 	case SECCOMP_FILTER_FLAG_WAIT_KILLABLE_RECV:
 		if (state.sup_flag_wait_kill < 0)
-			state.sup_flag_wait_kill = _sys_chk_flag_kernel(flag);
+			state.sup_flag_wait_kill = _sys_chk_flag_kernel(flag|SECCOMP_FILTER_FLAG_NEW_LISTENER);
```
pcmoore, line 315: Thanks for catching this, it matches up with how we handle the flag in `sys_filter_load()`, which is good, but I think we should add a check for the listener flag here and adjust the format a bit to fit better within an 80 char width.  Something like the patch below ...

``` diff
diff --git a/src/system.c b/src/system.c
index e205de7a..d0e3ecd5 100644
--- a/src/system.c
+++ b/src/system.c
@@ -311,8 +311,12 @@ int sys_chk_seccomp_flag(int flag)
                        state.sup_flag_tsync_esrch = _sys_chk_flag_kernel(flag);
                return state.sup_flag_tsync_esrch;
        case SECCOMP_FILTER_FLAG_WAIT_KILLABLE_RECV:
-               if (state.sup_flag_wait_kill < 0)
-                       state.sup_flag_wait_kill = _sys_chk_flag_kernel(flag);
+               /* kernel requires NEW_LISTENER with WAIT_KILLABLE_RECV */
+               flag |= SECCOMP_FILTER_FLAG_NEW_LISTENER;
+               sys_chk_seccomp_flag(SECCOMP_FILTER_FLAG_NEW_LISTENER);
+               if (!state.sup_flag_new_listener && !state.sup_flag_wait_kill)
+                       return 0;
+               state.sup_flag_wait_kill = _sys_chk_flag_kernel(flag);
                return state.sup_flag_wait_kill;
        }
```

alip: I had considered adding a check for the listener flag but I didn't as that is already checked before we hit this point. However, I suppose that is not necessarily always true and the check shouldn't hurt. I'll amend the patch, thanks!

alip: Please check out the latest fixup, I've edited your suggestion slightly for the `< 0` case but otherwise it's identical and covers the listener check.

pcmoore: I'm sorry for the delay, but this latest version looks good to me - thanks!

pcmoore (approved):
This looks good to me, what do you think @drakenclimber?  A note to whoever ends up merging this, squash the fixup patch into the first patch so there is only one commit in the tree.
```
```

drakenclimber (approved):

PR #424, opened 2023-12-16
BUG: avoid use of non-portable `which` in the tests

thesamesam:
which is not a standard POSIX utility, and indeed, each of these test scripts uses #!/bin/bash as its shebang, so we can use `type -P` which has the same behaviour as `which` for free.

(If the tests used POSIX shell, we could do `command -v`, its only caveat is that it'll pick up functions in the user's shell, which doesn't matter 99% of the time anyway.)

Distributions like Debian [0] and Gentoo [1] are looking to remove `which` from their base set of packages.

[0] https://lwn.net/Articles/874049/
[1] https://bugs.gentoo.org/646588

### DIFF
--- a/tests/38-basic-pfc_coverage.sh
+++ b/tests/38-basic-pfc_coverage.sh
@@ -18,7 +18,7 @@
 #
 function check_deps() {
 	[[ -z "$1" ]] && return
-	which "$1" >& /dev/null
+	type -P "$1" >& /dev/null
 	return $?
 }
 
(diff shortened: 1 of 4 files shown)


pcmoore:
Hi @thesamesam, first off, great branch name :)

Other than that, this patch looks good to me, and considering the trivial nature I'm going to go ahead and merge this into main via 47ca6441d62e41ba845ad8036c15f1154bd56b24.  @drakenclimber if you have any concerns/objections please let me know and we can fix things up, or revert this patch if needed.

Thanks all!

PR #433, opened 2024-06-22
scmp_bpf_sim: fix aliasing UB

thesamesam:
See https://github.com/seccomp/libseccomp/pull/425.

Punning sys_data_b between uint32_t* and struct* seccomp_data isn't legal, use memcpy to fix the testsuite with Clang 17.

Modern compilers recognise this idiom and optimise it out anyway.

### DIFF
--- a/tools/scmp_bpf_sim.c
+++ b/tools/scmp_bpf_sim.c
@@ -182,7 +182,8 @@ static void bpf_execute(const struct bpf_program *prg,
 		switch (code) {
 		case BPF_LD+BPF_W+BPF_ABS:
 			if (k < BPF_SYSCALL_MAX) {
-				uint32_t val = *((uint32_t *)&sys_data_b[k]);
+				uint32_t val;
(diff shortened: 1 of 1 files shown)


RossComputerGuy:
Still experiencing failures with LLVM 17.0.6 https://termbin.com/j2ey

thesamesam:
I can't reproduce that, I'm afraid. I get failures if I revert my patch, but it passes otherwise.

RossComputerGuy:
> Are you sure it was applied?

I am 100% the patch is applied, I'm using Nix and it would fail if the patch failed to apply. It said the patch was applied.

thesamesam:
No idea then, I'm afraid. You'll need to try debug it.

I suggest starting with object files from a good and bad tree and bisecting via those.

drakenclimber (approved):
Looks reasonable to me.  Thanks, @thesamesam 

thesamesam:
Thanks all!

PR #437, opened 2024-09-20
RFE: basic description of our release process

pcmoore:
We have maintainer docs regarding our release process, but nothing for normal users.  Let's fix that with a small section in our README.md.


drakenclimber (approved):

drakenclimber:
Good addition, @pcmoore.  I've added it to main in commit 9c4eb5eaf088.  Thanks!

PR #443, opened 2024-12-16
ci: bump ubuntu 24.04

t0b3:


.github/workflows/codeql-analysis.yml
```diff
@@ -11,7 +11,7 @@ on: ["push", "pull_request"]
 jobs:
   analyze:
     name: Analyze
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
pcmoore, line 14: I wonder if we should convert everything over to `ubuntu-latest` so we don't have to worry about chasing Ubuntu versions?

t0b3: there are basically two options
- pin os version ... thus chasing them (while controlling upgrades)
- use latest os ... hunting portability issues between os version upon upgrade (happening uncontrolled) :-(

nb: such new issues arose with ubuntu 24.04 (while i.e. 22.04 was straight)

I tend towards the first option while automating the check for version upgrades using dependabot :+1:

drakenclimber: > there are basically two options
> 
> * pin os version ... thus chasing them (while controlling upgrades)
> * use latest os ... hunting portability issues between os version upon upgrade (happening uncontrolled) :-(
> 
> nb: such new issues arose with ubuntu 24.04 (while i.e. 22.04 was straight)
> 
> I tend towards the first option while try to automating the check for version upgrades using dependabot 👍

I also have a slight preference toward the first option.  While not directly applicable to libseccomp, Ubuntu 24.04 now defaults to cgroup v2.  That's a big change for other projects, and it isn't one I would want to automatically pick up without thoroughly thinking through the ramifications.

t0b3: well let's go with the pinned os-version then :+1:

Makefile.am
```diff
@@ -22,7 +22,6 @@ CODE_COVERAGE_OUTPUT_FILE = libseccomp.lcov.info
 CODE_COVERAGE_OUTPUT_DIRECTORY = libseccomp.lcov.html.d
 CODE_COVERAGE_IGNORE_PATTERN = \
 	*/usr/include/* \
-	*/src/arch-syscall-check.c \
```
kolyshkin, line 25: Is this hunk relevant to this PR?

t0b3: @kolyshkin indeed it is see in https://github.com/t0b3/libseccomp/actions/runs/12359993960/job/34493901917#step:5:9488 coverage fails otherwise

kolyshkin (approved):
LGTM

PS According to https://github.com/actions/runner-images/issues/11101, Ubuntu 20.04 will be removed in April, and in March there will be a few "brown-out" dates/times when it won't work. So it makes sense to merge it now, or we'll have broken CI. Cc @drakenclimber @pcmoore

pcmoore (approved):
Looks good to me, you okay with this @drakenclimber?

`Acked-by: Paul Moore <paul@paul-moore.com>`

drakenclimber (approved):
LGTM

drakenclimber:
> Looks good to me, you okay with this @drakenclimber?
> 
> `Acked-by: Paul Moore <paul@paul-moore.com>`

I can merge it tomorrow if you don't get to it tonight.

pcmoore:
No worries, I'll take care of it today.

PR #62, opened 2016-12-26
RFE: update syscalls for Linux 4.9

justincormack:
- preadv2 and pwritev2
- pkey_mprotect, pkey_alloc, pkey_free

Signed-off-by: Justin Cormack <justin.cormack@docker.com>

fix #60

PR #452, opened 2025-01-25
BUG: remove the fuzzer from test 62-sim-arch_transactions

pcmoore:
We can't reliably run the bpf-sim-fuzz tests on tests which manipulate the filters arch/ABIs unless the filter is safe to run on all arch/ABIs, which is more or less impossible.  Remove the bpf-sim-fuzz test section in test #62 to work around this, just as we do with the other similar tests.


drakenclimber (approved):

drakenclimber:
Pushed to main in commit 7db46d72f13c.  Pushed to the release-2.6 branch in commit 2f0f3b0e9121.  Thanks, @pcmoore

PR #458, opened 2025-02-13
BUG: fix seccomp_export_bpf_mem out-of-bounds read

alyssais:
`*len` is the length of the destination buffer, but `program->blks` is probably not anywhere near that long.  It's already been checked above that `BPF_PGM_SIZE(program)` is less than or equal to `*len`, so that's the correct value to use here to avoid either reading or writing too much.

I noticed this because tests/11-basic-basic_errors started failing on musl after e797591 ("all: add seccomp_precompute() functionality").


drakenclimber (approved):

pcmoore:
Merged into main via e8dbc6b555fb936bdfb8ab86f9a45fda96a8b7a2, keeping open until the backport to release-2.6 is complete.

PR #459, opened 2025-02-18
BUG: fix strict aliasing UB in MurMur hash implementation.

Romain-Geissler-1A:
This was spotted when trying to upgrade the libseccomp fedora package to version 2.6.0 in fedora rawhide. It comes with gcc 15 and LTO enabled by default. When running the test 61-sim-transactions we get plenty of such errors in valgrind:

```
==265507== Use of uninitialised value of size 8
==265507==    at 0x4096AD: _hsh_add (gen_bpf.c:599)
==265507==    by 0x40A557: UnknownInlinedFun (gen_bpf.c:2016)
==265507==    by 0x40A557: gen_bpf_generate (gen_bpf.c:2341)
==265507==    by 0x400CDE: UnknownInlinedFun (db.c:2685)
==265507==    by 0x400CDE: UnknownInlinedFun (db.c:2682)
==265507==    by 0x400CDE: UnknownInlinedFun (api.c:756)
==265507==    by 0x400CDE: UnknownInlinedFun (util.c:162)
==265507==    by 0x400CDE: UnknownInlinedFun (util.c:153)
==265507==    by 0x400CDE: main (61-sim-transactions.c:128)
==265507==  Uninitialised value was created by a stack allocation
==265507==    at 0x409590: _hsh_add (gen_bpf.c:573)
```

[EDITED this part after the review of @thesamesam] Investigating this a bit, it seems that because of LTO the MurMur hash implementation is being inlined in _hsh_add. The two buffers data and blocks to point at the same underlying data, but via incompatible type, which is a strict aliasing violation. Instead, remove the getblock32 function and inline the copy with memcpy.

This is reproducible on a "fedora:rawhide" container (gcc 15) and using:
`export CFLAGS='-O2 -flto=auto -ffat-lto-objects -g'`

### DIFF
--- a/src/hash.c
+++ b/src/hash.c
@@ -12,14 +12,10 @@
  */
 
 #include <stdlib.h>
+#include <string.h>
 #include <inttypes.h>
 
 #include "hash.h"
-
-static inline uint32_t getblock32(const uint32_t *p, int i)
-{
-	return p[i];
-}
 
 static inline uint32_t rotl32(uint32_t x, int8_t r)
 {
@@ -41,7 +37,6 @@ static inline uint32_t fmix32(uint32_t h)
 uint32_t hash(const void *key, size_t length)
 {
 	const uint8_t *data = (const uint8_t *)key;
-	const uint32_t *blocks;
 	const uint8_t *tail;
 	const int nblocks = length / 4;
 	const uint32_t c1 = 0xcc9e2d51;
@@ -54,9 +49,8 @@ uint32_t hash(const void *key, size_t length)
 	uint32_t h1 = 0;
 
 	/* body */
-	blocks = (const uint32_t *)(data + nblocks * 4);
 	for(i = -nblocks; i; i++) {
-		k1 = getblock32(blocks, i);
+		memcpy(&k1, data + (nblocks + i) * sizeof(uint32_t), sizeof(uint32_t));
 
 		k1 *= c1;
 		k1 = rotl32(k1, 15);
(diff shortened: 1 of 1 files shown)


Romain-Geissler-1A:
Friendly ping @pcmoore ;)

src/hash.c
```diff
@@ -56,7 +52,7 @@ uint32_t hash(const void *key, size_t length)
 	/* body */
 	blocks = (const uint32_t *)(data + nblocks * 4);
```
thesamesam, line 53: The cast shouldn't be needed anymore (just inline it to the memcpy args)?

thesamesam:
>  The way we call getblock32 with the explicit cast to const uint32_t* is a strict aliasing violation.

It's accessing via a different (non-compatible) type, it's not about the cast.

src/hash.c
thesamesam: Thanks!

thesamesam (approved):
This looks good to me now, thanks!

`Reviewed-by: Sam James <sam@gentoo.org>`

pcmoore:
Pushed to main via 614530bc8b3c9f49aa59d7eaef4863b746504c23, keeping open until the backport is complete.

pcmoore:
Merged into release-2.6 via 84005ecc603fd0186188c4113452fd8e8a0c9bb3, thank you!

Romain-Geissler-1A:
Thanks !

Note that technically this may affect releases 2.5.x too, is the branch 2.5 still maintained ? If yes you may backport it there as well.

pcmoore:
Yes, release-2.5 is still currently maintained, but not for much longer and the bar for backporting to it is pretty high.

Unofficially, we tend to support v2.n until v2.(n+1).1 is released, but that isn't a hard rule.

PR #468, opened 2025-07-30
RFE: Add a Github Actions job to build libseccomp with clang

drakenclimber:
On other projects I've noticed that clang reports a slightly different list of warnings than gcc.  Add a Github Actions job to build libseccomp with clang.


pcmoore:
Merged the first patch in the PR (checkout version bump) via e7e633c28aed5333b185bfc0ad6f8d70b5fc20be.

pcmoore:
Merged via bbd421a3a67802e1f18d4fcb930f862e537e55ba, thanks!

PR #470, opened 2025-08-21
BUG: use notification structs from include/seccomp.h

pcmoore:
We don't want to duplicate struct definitions unless absolutely necessary, so rely on seccomp.h definitions inside system.h.  A very minor thing, but we should fix this at some point.

### DIFF
--- a/src/system.h
+++ b/src/system.h
@@ -156,28 +156,14 @@ typedef struct sock_filter bpf_instr_raw;
 #define SECCOMP_RET_LOG			0x7fc00000U
 #endif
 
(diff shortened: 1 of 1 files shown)


drakenclimber (approved):
LGTM.  I compared our current list of defines with what is in the kernel's seccomp.h, and I think you've defined the bare minimum in the case where those structs/#defines don't exist.

pcmoore:
Merged via dee17d04b0b602294a6859c8b36a6d07c0463691, thanks.

Issue #472, opened 2025-08-26
Test failure on x32


thesamesam:
This is with 2.6.0 with the test fixes on the branch cherry-picked:
```
 batch name: 60-sim-precompute
 test mode:  c
 test type:  bpf-sim
Test 60-sim-precompute%%001-00001 result:   FAILURE bpf_sim resulted in KILL
Test 60-sim-precompute%%001-00002 result:   FAILURE bpf_sim resulted in KILL
Test 60-sim-precompute%%001-00003 result:   FAILURE bpf_sim resulted in KILL
Test 60-sim-precompute%%001-00004 result:   FAILURE bpf_sim resulted in KILL
Test 60-sim-precompute%%001-00005 result:   FAILURE bpf_sim resulted in KILL
Test 60-sim-precompute%%001-00006 result:   FAILURE bpf_sim resulted in KILL
Test 60-sim-precompute%%001-00007 result:   FAILURE bpf_sim resulted in KILL
Test 60-sim-precompute%%001-00008 result:   FAILURE bpf_sim resulted in KILL
Test 60-sim-precompute%%001-00009 result:   FAILURE bpf_sim resulted in KILL
Test 60-sim-precompute%%001-00010 result:   FAILURE bpf_sim resulted in KILL
Test 60-sim-precompute%%001-00011 result:   FAILURE bpf_sim resulted in KILL
Test 60-sim-precompute%%002-00001 result:   SUCCESS
Test 60-sim-precompute%%003-00001 result:   SUCCESS
[...]
 batch name: 61-sim-transactions
 test mode:  c
 test type:  bpf-sim
Test 61-sim-transactions%%001-00001 result:   SUCCESS
Test 61-sim-transactions%%002-00001 result:   SUCCESS
Test 61-sim-transactions%%003-00001 result:   SUCCESS
Test 61-sim-transactions%%004-00001 result:   SUCCESS
Test 61-sim-transactions%%005-00001 result:   SUCCESS
Test 61-sim-transactions%%005-00002 result:   SUCCESS
Test 61-sim-transactions%%005-00003 result:   SUCCESS
Test 61-sim-transactions%%005-00004 result:   SUCCESS
Test 61-sim-transactions%%006-00001 result:   FAILURE bpf_sim resulted in KILL
Test 61-sim-transactions%%007-00001 result:   SUCCESS
Test 61-sim-transactions%%007-00002 result:   SUCCESS
[...]

PR #473, opened 2025-08-29
BUG: Skip tests 60 and 61 on x32

drakenclimber:
x32 syscalls require the X32_SYSCALL_BIT (0x4000 0000) to be set, and thus syscalls less than that are invalid on the x32 architecture.  Tests 60 and 61 utilize syscalls less than 0x4000 0000, so skip those tests for the x32 arch.

Fixes Github Issue #472.

Reported-by: Sam James <sam@gentoo.org>


pcmoore:
Merged via 9b9ea8e7a173b96e59fb21e8d461365110e7b4ef with some indent fixes on the test headers for proper alignment, thanks!

PR #478, opened 2026-01-08
BUG: fix shebang for python3

### LINKED ISSUE #477: Use `python3` hashbang in scripts
As per https://peps.python.org/pep-0394/, the naming of the Python interpreter on Linux is a bit of a mess.

Notably, whilst all distros have `python3`, not all distros have `python`.

Thus it seems sensible to change all the hashbangs to `#! /usr/bin/env python3` as they _are_ python3 scripts and who knows what happens to `python` when python4 is released, if that ever happens!

RocketMaDev:
This is a commit apply fix from #440 since the old fix was not completed and stale. Just `sed -i 's|#!/usr/bin/env python|#!/usr/bin/env python3|' tests/*.py` could fix this problem.

Fix #477, close #440.

### DIFF
--- a/src/python/setup.py
+++ b/src/python/setup.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 #
 # Enhanced Seccomp Library Python Module Build Script


pcmoore:
Thanks for picking this up @RocketMaDev.  This looks good to me, so I'm going to go ahead and merge this into the main branch via 18e4955cd0d92de243624eaebdbb18aa921b4b18, but leave this PR open so we can discuss backporting to the release-2.6 branch.  I will note that I manually performed the /python/python3/ change in test 63 as it was just merged before this PR.

What do you think @drakenclimber, should we backport this for v2.6.z?  I'm inclined to backport it.

pcmoore:
I was talking with @drakenclimber outside this issue and he agreed we should do the backport, so I've done just that at 6abe14d1645e24ce3d30637204747d93f6909356.

Thanks again everyone!

PR #484, opened 2026-04-06
RFE: add versioned symbol support

gordonmessmer:
In order to improve reliable dependency resolution, I would like to add versioned symbols to the libraries that do not include them today.

This change is intended to support Fedora's recommendation that shared libraries provide versioned symbols: https://docs.fedoraproject.org/en-US/packaging-guidelines/C_and_C++/#_versioned_symbols

Please double-check the symbol files. If any symbols have been added very recently, they might not be in the symbol maps I've generated.

### DIFF
--- a/src/libseccomp.map
+++ b/src/libseccomp.map
@@ -1,0 +1,44 @@
+/* Avoid modifying a symbol set after it has been released
+   When adding features in a new release, add a new set
+   Removing features is a breaking change */
+LIBSECCOMP_2.7 {
+  global:
+    seccomp_api_get;
+    seccomp_api_set;
+    seccomp_arch_add;
+    seccomp_arch_exist;
+    seccomp_arch_native;
+    seccomp_arch_remove;
+    seccomp_arch_resolve_name;
+    seccomp_attr_get;
+    seccomp_attr_set;
+    seccomp_export_bpf;
+    seccomp_export_bpf_mem;
+    seccomp_export_pfc;
+    seccomp_init;
+    seccomp_load;
+    seccomp_merge;
+    seccomp_notify_alloc;
+    seccomp_notify_addfd;
+    seccomp_notify_fd;
+    seccomp_notify_free;
+    seccomp_notify_id_valid;
+    seccomp_notify_receive;
+    seccomp_notify_respond;
+    seccomp_precompute;
+    seccomp_release;
+    seccomp_reset;
+    seccomp_rule_add;
+    seccomp_rule_add_array;
+    seccomp_rule_add_exact;
+    seccomp_rule_add_exact_array;
+    seccomp_syscall_priority;
+    seccomp_syscall_resolve_name;
+    seccomp_syscall_resolve_name_arch;
+    seccomp_syscall_resolve_name_rewrite;
+    seccomp_syscall_resolve_num_arch;
+    seccomp_transaction_commit;
+    seccomp_transaction_reject;
+    seccomp_transaction_start;
+    seccomp_version;
+};
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -49,7 +49,8 @@ SOURCES_ALL = \
 
 EXTRA_DIST = \
 	arch-syscall-validate arch-syscall-check \
-	arch-gperf-generate syscalls.csv syscalls.perf.template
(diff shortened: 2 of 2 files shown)


pcmoore:
Hi @gordonmessmer, thanks for the patch!  Considering the patch, it would be helpful to provide a description of the work in the patch's description.  We also generally expect a subject prefix, e.g. "build: add versioned symbol support" for patches so it is easier to filter through patches when looking at the output from `git log`.

m4/ld-version-script.m4
```diff
@@ -0,0 +1,50 @@
...
+dnl gives unlimited permission to copy and/or distribute it,
+dnl with or without modifications, as long as this notice is preserved.
+dnl This file is offered as-is, without any warranty.
+
+dnl From Simon Josefsson
+
+# FIXME: The test below returns a false positive for mingw
+# cross-compiles, 'local:' statements does not reduce number of
+# exported symbols in a DLL.  Use --disable-ld-version-script to work
+# around the problem.
```
pcmoore: Is there a fix for this, or is this a known problem that is expected to have a solution at some later date?

m4/ld-version-script.m4
```diff
@@ -0,0 +1,50 @@
...
+
+# gl_LD_VERSION_SCRIPT
+# --------------------
+# Check if LD supports linker scripts, and define automake conditional
```
drakenclimber: Are there any (modern) versions of LD that don't support linker scripts?

m4/ld-version-script.m4
gordonmessmer: That comment is 17 years old: https://github.com/coreutils/gnulib/blame/master/m4/ld-version-script.m4

I don't really work with mingw, so I don't know if that was a defect in its ld, or if it's a limitation imposed by the target format.

I can research the question further if it's important.

m4/ld-version-script.m4
gordonmessmer: As far as I know, they are supported by GNU ld, LLVM's lld, mold, and wild.

drakenclimber: If we think all (modern) versions of LD support linker scripts, do we even need this *.m4 file?  I work on another upstream project, [libcgroup](https://github.com/libcgroup/libcgroup), and we've used linker scripts for years.  We don't have any checks for versions of LD that don't support linker scripts.  No one has complained yet :)

gordonmessmer: I've removed those tests in the new patch. :)

gordonmessmer:
Checking in: Please let me know if there are any remaining changes you'd like to see in this PR.

src/libseccomp.map
```diff
@@ -0,0 +1,43 @@
+/* Avoid modifying a symbol set after it has been released
+   When adding features in a new release, add a new set
+   Removing features is a breaking change */
+LIBSECCOMP_1.0 {
```
drakenclimber: I would vote to make this `LIBSECCOMP_2.7` since this patchset would be part of that release.

Then the symbol versions would be the same as the library versions.  Seems like that would be easier to track

src/libseccomp.map
```diff
@@ -0,0 +1,43 @@
...
+    seccomp_arch_resolve_name;
+    seccomp_attr_get;
+    seccomp_attr_set;
+    seccomp_export_bpf;
+    seccomp_export_bpf_mem;
+    seccomp_export_pfc;
+    seccomp_init;
+    seccomp_load;
+    seccomp_merge;
+    seccomp_notify_alloc;
```
drakenclimber, line 21: Looks like we missed `seccomp_notify_addfd()`

drakenclimber (commented):
The patch looks good to me.  I have a couple small change requests, and then I think it's good to go from my end.

Also, I tested this patch against `make distcheck` and `make dist-gzip` and unzipped the tar file.  Everything worked as expected.

Thanks, @gordonmessmer

src/libseccomp.map
gordonmessmer: Thanks for catching that!

src/libseccomp.map
gordonmessmer: Makes sense. I've updated the commit.

Issue #480, opened 2026-02-15
Please clarify on how to update syscalls.csv


glaubitz:
I have started working on Alpha support as [support for SECCOMP_FILTER is now being worked on for Alpha](https://lore.kernel.org/all/20260205133049.526-2-linmag7@gmail.com/).

I have added the usual stuff [to add Alpha support](https://github.com/glaubitz/libseccomp/tree/alpha-support) and wanted to update `syscalls.csv` now to add the syscalls for Alpha.

However, it seems that I'm no longer able to do that since the scripts for that were rewritten and it's not clear to me how the scripts are supposed to be used:

```
glaubitz@z6:..upstream/libseccomp> ./src/arch-update-syscalls-csv.py -k /home/glaubitz/upstream/linux -c src/syscalls.csv -V 6.18 -d datapath/
Updating src/syscalls.csv version table for kernel 6.18
Traceback (most recent call last):
  File "/home/glaubitz/upstream/libseccomp/./src/arch-update-syscalls-csv.py", line 330, in <module>
    main(args)
    ~~~~^^^^^^
  File "/home/glaubitz/upstream/libseccomp/./src/arch-update-syscalls-csv.py", line 320, in main
    syscalls = update_syscalls_dict(args, columns, syscalls, kver)
  File "/home/glaubitz/upstream/libseccomp/./src/arch-update-syscalls-csv.py", line 224, in update_syscalls_dict
    with open(table_path, 'r') as tblf:
         ~~~~^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'datapath/tables-6.18/syscalls-i386'
glaubitz@z6:..upstream/libseccomp>
```

In particular, the script is now asking for a `datapath` and I have no clue what's that supposed to be. I have skimmed through the documentation and source code, but it doesn't become clear.

Can you please add instructions to `README.md` on how to create or update `syscalls.csv`?

Thanks!

drakenclimber:
I can put something together.

And sorry for being AWOL lately.  I'm not certain what my future holds, but I'm hoping to be a little more present upstream over the next few months.

drakenclimber:
[...]

PR #485, opened 2026-04-14
RFE: Update syscalls.csv to kernel version v7.0-rc7

drakenclimber:
In this patchset:
* Update syscalls.csv to the latest kernel version, v7.0-rc7
* Document the process to update syscalls.csv
* Various minor improvements to the tools to build syscalls.csv
* Remove 32-bit s390 since it's no longer supported

For the libseccomp v2.6.x releases, I think we won't want to backport the s390-removal patch.

Here's a comparison of the HEAD syscalls.csv (with the s390 columns removed) and the proposed syscalls.csv.
https://www.textcompare.org/csv/?id=69dead128329b015e57a0320


drakenclimber:
Fixes: #480

drakenclimber:
~~Before we merge this patchset, we should probably also add this patch that sets the kernel versions for all existing syscalls.~~

~~https://github.com/seccomp/libseccomp/pull/457/changes/031024addec2fd77ff88b92ac2e8b68b56e9b746~~

~~I'll add it to this patchset.~~

Changed my mind. See the [comment](https://github.com/seccomp/libseccomp/pull/485#issuecomment-4263437378) below

drakenclimber:
v2 changes:
* Removed the patch that removes s390.  To minimize changes between the main and the release-2.6 branches, let's merge that patch after cutting release v2.6.1.  I'll add the s390 patch to pull request #457.
* Because S390 remains in this patchset, I had to add a couple more `__PNR` definitions

include/seccomp-syscalls.h
```diff
@@ -283,6 +283,9 @@
 #define __PNR_getpagesize			-10249
 #define __PNR_riscv_hwprobe			-10250
 #define __PNR_uretprobe				-10251
+#define __PNR_uprobe				-10252
+#define __PNR_rseq_slice_yield			-10253
```
pcmoore, line 287: Based on the macros below it looks like `__PNR_rseq_slice_yield` isn't necessary as `__NR_rseq_slice_yield` is defined for all arches/ABIs?

include/seccomp-syscalls.h
```diff
@@ -283,6 +283,9 @@
...
+#define __PNR_listns				-10254
```
pcmoore, line 288: Same here, it looks like `__NR_listns` is defined for all arches/ABIs?  If not, we need to so the __SNR to __NR or __PNR macro dance down below.

CONTRIBUTING.md
```diff
@@ -64,6 +64,105 @@ base, and can be enabled via the "--enable-code-coverage" configure flag and
...
+
+1. Install dependencies
+
+   In addition to the normal libseccomp package dependencies, the following
+   packages must also be installed:
+   ```
+   apt install libc6-dev-x32
+   ```
+
+1. Download source packages
```
pcmoore, line 79: All of the list items are numbered "1." :)

CONTRIBUTING.md
```diff
@@ -64,6 +64,105 @@ base, and can be enabled via the "--enable-code-coverage" configure flag and
 the "check-code-coverage" make target.  Additional details on generating code
 coverage information can be found in the .travis.yml file.
 
+## How to Update the syscalls.csv Table
+
+*** NOTE - This currently can only be done on Ubuntu ***
```
pcmoore, line 69: Is this just because of the package dependencies in step 1?  If so, perhaps we add unordered sub-list there for various distros, e.g. Fedora/RHEL, Arch, etc.

pcmoore:
> v2 changes:
> 
> * Removed the patch that removes s390.  To minimize changes between the main and the release-2.6 branches, let's merge that patch after cutting release v2.6.1.  I'll add the s390 patch to pull request [RFE: Add support for maximum supported kernel version #457](https://github.com/seccomp/libseccomp/pull/457).
> * Because S390 remains in this patchset, I had to add a couple more `__PNR` definitions

I think we need to keep S390 around for a few more years since it was only just removed in Linux v6.19.  I'll add a GH issue so we don't forget about it, but I suspect we may even want to keep it around for the v2.7.z stream too.

The good news is that from our perspective it's just a matter of PNR'ing any new syscall that comes along for S390.

include/seccomp-syscalls.h
drakenclimber: 32-bit s390 isn't getting updates, so it needs `__PNR_rseq_slice_yield`

include/seccomp-syscalls.h
drakenclimber: Same comment as above.  32-bit s390 needs it.

This is one reason I was tempted to ditch s390 now.

The other reason is that [syscalls table tool](https://github.com/hrw/syscalls-table) has generated some different syscall numbers from libseccomp's `arch-syscall-validate` for 32-bit s390.  This made me nervous

CONTRIBUTING.md
drakenclimber: This is intentional.  Github's Markdown parser converts them into correct numbers when rendered.

https://github.com/drakenclimber/libseccomp/blob/v7.0-rc7/CONTRIBUTING.md#how-to-update-the-syscallscsv-table

CONTRIBUTING.md
drakenclimber: I had troubles generating the correct syscall numbers for x32 on non-Ubuntu distros.   I tried installing various 32-bit packages, but I couldn't get it to work.

@hrw - can you successfully generate x32 syscalls on Red Hat, Fedora, Oracle Linux, etc.?

CONTRIBUTING.md
drakenclimber: With that said, if we think people are going to read the markdown file directly in vim/emacs/whatever, I can properly number the list.

CONTRIBUTING.md
hrw: My AArch64 system runs Fedora 44. Regenerated x32 tables just fine.

drakenclimber: Glad to hear that.  Do you know what 32-bit package(s) you installed?  I wasn't able to find the right ones, and as I pointed out, my x32 syscalls were wrong.  Thanks!

hrw:
```
$ rpm -qa|grep glibc
glibc-2.43-3.fc44.aarch64
glibc-common-2.43-3.fc44.aarch64
glibc-gconv-extra-2.43-3.fc44.aarch64
glibc-all-langpacks-2.43-3.fc44.aarch64
glibc-devel-2.43-3.fc44.aarch64
glibc-static-2.43-3.fc44.aarch64
```

That's all I have installed.

drakenclimber: Hmmm... My Fedora41 x86 box behaves quite a bit differently.

```
❯ cat /etc/fedora-release
Fedora release 41 (Forty One)

❯ rpm -qa|grep glibc
glibc-common-2.40-28.fc41.x86_64
glibc-all-langpacks-2.40-28.fc41.x86_64
glibc-gconv-extra-2.40-28.fc41.x86_64
glibc-2.40-28.fc41.x86_64
glibc-devel-2.40-28.fc41.x86_64
glibc-static-2.40-28.fc41.x86_64
```

When I run `scripts/update-tables.sh ...` on my Fedora box, a lot of the x32 syscalls get (erroneously?) populated with version numbers.  For example, lstat64 becomes syscall 196.

Here's a comparison of building the tables on my Ubuntu24 vs my Fedora41 machine.

https://www.textcompare.org/csv/?id=69f0fe048329b015e57a1c02

drakenclimber: I created a pull request in the syscalls-table repo.  Let's hold off on merging this patchset until we get that question/issue resolved.

https://github.com/hrw/syscalls-table/pull/138

drakenclimber: Thanks for the quick review and merging my pull request, @hrw.  This pull request is unblocked now.

PR #487, opened 2026-05-12
RFE: update syscall table to Linux v7.1.0-rc4

drakenclimber:
Update the syscall table to Linux v7.1.0-rc4.

    This table was updated by running:
    $ ./arch-syscall-validate -c syscalls.csv <kernel_directory>

    A few notes:
    * Kernel commit 8e0b986c59c6 ("s390: Remove compat support") removed
      31-bit s390 support.  Thus, any syscalls added to the kernel after
      this commit are not supported by 31-bit s390
    * On parisc64, _llseek() changed from syscall 140 to PNR. Kernel
      commit da3680f564bd ("parisc: _llseek syscall is only available
      for 32-bit userspace") moved this syscall to 32-bit parisc only.
    * The following syscalls were added:
      file_getattr(), file_setattr(), listns(), open_tree_attr(),
      rseq_slice_yield(), and uprobe()


pcmoore:
This looks good to me, but while backporting it to the release-2.6 branch I realized that we can't keep fixing up the s390 entries by hand, we'll go crazy.  Let me see if I can come up with something ...

pcmoore:
Okay, see #490 for a possible s390 solution.  I will note that the changes in #490 resulted in a different syscall table than this PR, but on quick glance it appeared to be correct (this PR looks to assign syscall numbers to syscalls beyond v6.18?).

drakenclimber:
> Okay, see #490 for a possible s390 solution. I will note that the changes in #490 resulted in a different syscall table than this PR, but on quick glance it appeared to be correct (this PR looks to assign syscall numbers to syscalls beyond v6.18?).

You're correct.  I should have PNR'ed syscalls newer than v6.18, but I didn't.  I'll fix it shortly.  Good catch.

drakenclimber:
I pushed a v2 of this patchset.  Changes include:
* PNR any 31-bit s390 syscall that is newer than v6.18
* In `arch-syscall-validate` ignore comment lines from ABI CSVs

pcmoore:
Merged into the main branch via 7e95f5b92c8fb5f2a21ef388111450773cb7d06a and the release-2.6 branch via fea23415b090472487279843918e8d60f977acd1, thanks!

PR #488, opened 2026-05-18
BUG: check direct syscall before multiplexed pseudo-syscall

nikita-dubrovskii:
Docker 29.4.2 removed socketcall(2) from the default seccomp profile. On s390x, this broke socket operations:
```
  # strace curl icanhazip.com
  socket(AF_UNIX, SOCK_STREAM, 0) = -1 ENOSYS (Function not implemented)

  # scmp_sys_resolver -a s390x socket
  -101

  # ausyscall s390x socket
  socket             359
```
The abi_syscall_resolve_name_munge() function was returning __PNR_socket (-101) instead of checking if arch implements socket(2) directly (359).

Fix by checking arch->syscall_resolve_name_raw() first, only falling back to multiplexed pseudo-syscalls if the direct implementation doesn't exist.

Affects socket and IPC syscalls on architectures with direct implementations (s390x, aarch64, etc).

### DIFF
--- a/src/syscalls.c
+++ b/src/syscalls.c
@@ -42,6 +42,13 @@
 int abi_syscall_resolve_name_munge(const struct arch_def *arch,
 				   const char *name)
 {
+	int sys = __NR_SCMP_ERROR;
+
+	/* Check if arch implements syscall directly and return its number */
+	sys = arch->syscall_resolve_name_raw(name);
+	if (sys != __NR_SCMP_ERROR) {
+		return sys;
+	}
 
 #define _ABI_SYSCALL_RES_NAME_CHK(NAME) \
 	if (!strcmp(name, #NAME)) return __PNR_##NAME;
@@ -79,7 +86,7 @@ int abi_syscall_resolve_name_munge(const struct arch_def *arch,
 	_ABI_SYSCALL_RES_NAME_CHK(shmget)
 	_ABI_SYSCALL_RES_NAME_CHK(shmctl)
 
-	return arch->syscall_resolve_name_raw(name);
+	return __NR_SCMP_ERROR;
 }
 
 /**
--- a/tests/30-sim-socket_syscalls.tests
+++ b/tests/30-sim-socket_syscalls.tests
@@ -44,9 +44,10 @@ test type: bpf-sim
 30-sim-socket_syscalls	+sh				352		0		1		2	N	N	N	ALLOW
 # direct syscalls
 30-sim-socket_syscalls	+x86,+ppc64le,+mipsel,+sh	accept		5		N		N	N	N	N	ALLOW
-30-sim-socket_syscalls	+x86,+ppc64le,+mipsel,+sh	accept		0		1		2	N	N	N	KILL
+30-sim-socket_syscalls	+x86				accept		0		1		2	N	N	N	KILL
+30-sim-socket_syscalls	+ppc64le,+mipsel,+sh		accept		0		1		2	N	N	N	ALLOW
 30-sim-socket_syscalls	+x86,+ppc64le,+mipsel,+sh	accept4		18		1		2	N	N	N	ALLOW
-30-sim-socket_syscalls	+x86,+ppc64le,+mipsel,+sh	accept4		0		1		2	N	N	N	KILL
+30-sim-socket_syscalls	+x86,+ppc64le,+mipsel,+sh	accept4		0		1		2	N	N	N	ALLOW
 30-sim-socket_syscalls	+x86_64				socket		0		1		2	N	N	N	ALLOW
 30-sim-socket_syscalls	+x86_64				connect		0		1		2	N	N	N	ALLOW
 30-sim-socket_syscalls	+x86_64				accept4		0		1		2	N	N	N	ALLOW
--- a/tests/33-sim-socket_syscalls_be.tests
+++ b/tests/33-sim-socket_syscalls_be.tests
@@ -21,9 +21,10 @@ test type: bpf-sim
 33-sim-socket_syscalls_be	+s390,+s390x		373		0		1		2	N	N	N	ALLOW
 33-sim-socket_syscalls_be	+ppc			338		0		1		2	N	N	N	ALLOW
 33-sim-socket_syscalls_be	+s390,+s390x,+ppc	accept		5		N		N	N	N	N	ALLOW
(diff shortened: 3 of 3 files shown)


nikita-dubrovskii:
This won't fix Docker itself, but at least the tools will return the correct values:
```
[zukku@a3elp43 libseccomp]$ scmp_sys_resolver socket
-101

$ LD_PRELOAD=./src/.libs/libseccomp.so scmp_sys_resolver socket
359

```

pcmoore:
We need to be careful with changes like this as it is changing the behavior of the library in a way that might not be friendly to all of the users.

pcmoore:
Marking this as a BUG for now, but until we understand the impact of a change like this we need to take the classification with a grain of salt ...

drakenclimber:
I think having libseccomp use the "standard" syscall before the multiplexed pseudo-syscall is a good discussion to have, but as @pcmoore pointed out, the ramifications could be large.  We would definitely have to proceed with caution if we were to make such a change.

With that said, is this still urgent?  It looks like Docker v29.4.3 [1] fixed this.  Correct?

[1] https://github.com/moby/moby/discussions/52555

nikita-dubrovskii:
> With that said, is this still urgent? It looks like Docker v29.4.3 [1] fixed this. Correct?

Yes, it should be working now.


> I think having libseccomp use the "standard" syscall before the multiplexed pseudo-syscall is a good discussion to have, but as @pcmoore pointed out, the ramifications could be large.

After checking git history, i have an impression, that current behaviour was implemented as a [workaround](https://github.com/seccomp/libseccomp/commit/66282c31034e2bb442bd9dc862d6d814f0df2f98) many years ago. Than several updates like [this](https://github.com/seccomp/libseccomp/commit/be65b26b67099be2b2b4890d736dbd1ad15adf36#diff-8662fa81ac722625889dc2b5c8524415a94cbec8f65f170fc0ee4585b8cbd176R56-R165) with some `mux`/`demux`, and finally [consolidation](https://github.com/seccomp/libseccomp/commit/17cbd2c253ce63e5e9e3cec867ff58efbe8b5fdc#diff-6542870a47f72a402f79386a7fd3b988371e7abeb220ec0166832e371cc63a51R331) moved all logic to centralized functions. But! Bug was still there - instead of returning direct syscall, code always returned multiplexed. 

The code DID work:
1. Name resolution (buggy): `abi_syscall_resolve_name_munge("socket")` returned `__PNR_socket` (-101) instead of 359
2. Dual-rule generation in `abi_rule_add()` [logic](https://github.com/seccomp/libseccomp/blob/main/src/syscalls.c#L400):
    - one for [direct syscall](https://github.com/seccomp/libseccomp/blob/main/src/syscalls.c#L450)
    - one for [multiplexed](https://github.com/seccomp/libseccomp/blob/main/src/syscalls.c#L438)
   
This fix doesn't break rule generation, as `abi_rule_add()` works correctly regardless of whether you start with the direct syscall (359) or pseudo-syscall (-101).

This is my understanding.

drakenclimber:
> The code DID work:
> 
> 1. Name resolution (buggy): `abi_syscall_resolve_name_munge("socket")` returned `__PNR_socket` (-101) instead of 359
> 2. Dual-rule generation in `abi_rule_add()` [logic](https://github.com/seccomp/libseccomp/blob/main/src/syscalls.c#L400):
>    
>    * one for [direct syscall](https://github.com/seccomp/libseccomp/blob/main/src/syscalls.c#L450)
>    * one for [multiplexed](https://github.com/seccomp/libseccomp/blob/main/src/syscalls.c#L438)
> 
> This fix doesn't break rule generation, as `abi_rule_add()` works correctly regardless of whether you start with the direct syscall (359) or pseudo-syscall (-101).
> 
> This is my understanding.

Thanks.  Your research is thorough and spot on.

The worry I have is that this will change the default multiplexing behavior for _all_ users on _all_ architectures.  I don't know what the fallout would be in that case.

nikita-dubrovskii:
> The worry I have is that this will change the default multiplexing behavior for all users on all architectures. I don't know what the fallout would be in that case.

It won't. I've made a test, here are outputs without and with the fix.

Before:
```
$ ./65-s390x-syscalls 
FAILED: seccomp_syscall_resolve_name_arch(SCMP_ARCH_S390X, "socket") expected 359, got -101
PASSED: seccomp_syscall_resolve_num_arch(SCMP_ARCH_S390X, 359) expected "socket", got socket
PASSED: seccomp_syscall_resolve_name_arch(SCMP_ARCH_S390X, "accept") expected -105, got -105
FAILED: seccomp_syscall_resolve_name_arch(SCMP_ARCH_S390X, "accept4") expected 364, got -118
PASSED: seccomp_syscall_resolve_num_arch(SCMP_ARCH_S390X, 364) expected "accept4", got accept4
PASSED: seccomp_syscall_resolve_name_arch(SCMP_ARCH_S390X, "socketcall") expected 102, got 102
PASSED: _abi_syscall_demux(&arch_def_s390x, -101) expected 359, got 359
PASSED: _abi_syscall_mux(&arch_def_s390x, 359) expected -101, got -101
PASSED: _abi_syscall_demux(&arch_def_s390x, -105) expected -105, got -105
PASSED: _abi_syscall_mux(&arch_def_s390x, 364) expected -118, got -118
```

After:
```
$ ./65-s390x-syscalls 
PASSED: seccomp_syscall_resolve_name_arch(SCMP_ARCH_S390X, "socket") expected 359, got 359
PASSED: seccomp_syscall_resolve_num_arch(SCMP_ARCH_S390X, 359) expected "socket", got socket
PASSED: seccomp_syscall_resolve_name_arch(SCMP_ARCH_S390X, "accept") expected -105, got -105
PASSED: seccomp_syscall_resolve_name_arch(SCMP_ARCH_S390X, "accept4") expected 364, got 364
PASSED: seccomp_syscall_resolve_num_arch(SCMP_ARCH_S390X, 364) expected "accept4", got accept4
PASSED: seccomp_syscall_resolve_name_arch(SCMP_ARCH_S390X, "socketcall") expected 102, got 102
PASSED: _abi_syscall_demux(&arch_def_s390x, -101) expected 359, got 359
PASSED: _abi_syscall_mux(&arch_def_s390x, 359) expected -101, got -101
PASSED: _abi_syscall_demux(&arch_def_s390x, -105) expected -105, got -105
PASSED: _abi_syscall_mux(&arch_def_s390x, 364) expected -118, got -118
```

[Uploading 0001-tests-add-tests-for-syscalls.h-api.patch…]()

